PR #13, opened 2015-01-27
Detect mkv without .slice method

shinnn:
When the argument is Uint8Array, it throws an error because Uint8Array doesn't have `slice` method.


kevva:
Same goes for `epub`.

PR #14, opened 2015-02-03
add woff,woff2,eot

junmer:


index.js
```diff
@@ -223,5 +223,29 @@ module.exports = function (buf) {
 		};
 	}
 
+	// http://www.w3.org/TR/WOFF/#WOFFHeader
```
junmer: yep, i removed it :blush:

PR #18, opened 2015-03-02
Add support for `flv` type

arthurvr:


arthurvr:
@sindresorhus I recently needed to build [is-postscript](https://github.com/arthurvr/is-postscript). You want it in this repo as well?

sindresorhus:
@arthurvr Hm, is it safe, it looks a bit short, just checking two bytes? But sure.

arthurvr:
> Hm, is it safe, it looks a bit short, just checking two bytes?

Yeah, I looked weird to me as well, but it's considered its magic number, so...

PR coming.

PR #19, opened 2015-03-07
Add suppport for `ps` type

arthurvr:
postscript detection :cake:

https://github.com/sindresorhus/file-type/pull/18#issuecomment-77688922

### DIFF
--- a/index.js
+++ b/index.js
@@ -272,5 +272,12 @@ module.exports = function (buf) {
 		};
 	}
 
+	if (buf[0] === 0x25 && buf[1] === 0x21) {
+		return {
+			ext: 'ps',
+			mime: 'application/postscript'
+		};
+	}
+
 	return null;
 };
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "rar",
     "gz",
     "bz2",
+    "ps",
```
sindresorhus: I know this is crazy, but can you put it at the end so everything is in the same order?

index.js
```diff
@@ -4,6 +4,13 @@ module.exports = function (buf) {
 		return null;
 	}
 
+	if (buf[0] === 0x25 && buf[1] === 0x21) {
+		return {
+			ext: 'ps',
+			mime: 'application/postscript'
+		};
+	}
+
```
sindresorhus: Can you put this test at the end too?

arthurvr: done

package.json
arthurvr: done

PR #23, opened 2015-06-04
Added m4v support.

nleclerc:
I've added m4v support, as generated by ffmpeg.
I've also modified m4a test since they are very similar.


test.js
```diff
@@ -25,6 +25,7 @@ var types = [
 	'bz2',
 	'7z',
 	'mp4',
+	'm4v',
```
kevva: Keep the same order here as in `index.js`.

nleclerc:
Done.

arthurvr:
Also update the readme.

nleclerc:
oh yeah, sorry.

sindresorhus:
I thought `m4v` was the same as `mp4`. Do you have any docs on this?

nleclerc:
I thought the same but i got errors because some of my mp4 files where detected as m4a.
I did a bit of digging and if the formats are mostly the same, m4v is an apple derivative that may contain drm.

I did chose to use m4v because the file does explicitely contain an M4V string.

http://www.winxdvd.com/resource/why-handbrake-saves-dvd-to-m4v-not-mp4.htm
http://www.macxdvd.com/mac-dvd-video-converter-how-to/m4v-vs-mp4.htm
http://en.wikipedia.org/wiki/M4V

PR #25, opened 2015-07-19
Add support for MIDI type

trexnix:
Hi, I would like to add support for MIDI type. One problem, however, its extensions can fall into either `mid` or `midi`. I also found out that many file types have the same file magic numbers too, `docx`, `xlsx`, `pptx`, for example. Do we need to cover those cases?


arthurvr:
> I also found out that many file types have the same file magic numbers too, docx, xlsx, pptx, for example. Do we need to cover those cases?

I'd rather exclude a type than having it reporting false positives, honestly.

sindresorhus:
Can you link to some documentation on this and the docx,etc magic numbers? There might be other thing in the MIDI file we could check.

> One problem, however, its extensions can fall into either mid or midi

That applies to a lot of file types, and why we supply the mime type too.

trexnix:
Glad to hear back.

Here are some documentation which all say MIDI's file magic numbers is `4D 54 68 64`: [Link 1](https://en.wikipedia.org/wiki/List_of_file_signatures), [Link 2](https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Magic_numbers_in_files), [Link 3](https://asecuritysite.com/forensics/magic) and this promising [file signatures table by Gary Kessler](http://www.garykessler.net/library/file_sigs.html). I think `mid` to `midi` is just like `htm` to `html`, or [`jpg` to `jpeg`](http://stackoverflow.com/questions/23424399/jpg-vs-jpeg-image-formats) there is nothing difference between their binary contents thus extensions are just a matter of taste. Oh hey, this case is exactly the same as `jpg` and `jpeg`, I wonder how did you guys pick one over the other?

Above resources also refer to `docx`, `xlsx`, etc, but in an inconsistency way, they have some parts in common though.

Sorry in advance for my English.

sindresorhus:
> I mean, the return value say it's jpg, not jpeg.

We pick the most common one.

trexnix:
From my point of view, `.mid` is more popular than `.midi`. If you find this file type not a commonly used one which is not suitable for the purpose of the project then you could drop it. I'm totally okay with that.

sindresorhus:
I'm fine with `.mid`. I'm traveling at the moment, but will look more into this PR in the weekend.

sindresorhus:
> Above resources also refer to docx, xlsx, etc, but in an inconsistency way, they have some parts in common though.

I still don't see how `docx`/`xlsx` are conflicting with `mid`? According to https://en.wikipedia.org/wiki/List_of_file_signatures they just have the ZIP magic number.

trexnix:
Sorry, I hadn't been clear at the first place and you got me wrong. What I was trying to say is `mid` and `midi` have same file signature, I also found out that `docx` and `xlsx` have  same file signature too. I just point out another case in which difference file extensions have same file signature and wonder how do we solve those cases, not that `mid` and `docx`, `xlsx` are conflicting. Hopefully you get what I mean :crying_cat_face:

sindresorhus:
Thanks. That makes it clearer. Available in 2.10.0 ;)

trexnix:
Nice. Thanks :)

PR #27, opened 2015-07-21
Add CR2 support

gillstrom:
Added support for CR2. The first bytes are the same as TIFF, so it had to be placed before.

The extension is capitalized (`.CR2`) importing the photos from my camera, so I kept it that way. Should it be in lowercase?

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,11 +32,12 @@ module.exports = function (buf) {
 		};
 	}
 
+	// needs to be before `tif` check
 	if (((buf[0] === 0x49 && buf[1] === 0x49 && buf[2] === 0x2A && buf[3] === 0x0) || (buf[0] === 0x4D && buf[1] === 0x4D && buf[2] === 0x0 && buf[3] === 0x2A)) && buf[8] === 0x43 && buf[9] === 0x52) {
 		return {
-			ext: 'CR2',
+			ext: 'cr2',
 			mime: 'image/x-canon-cr2'
-		}
+		};
 	}
 
 	if ((buf[0] === 0x49 && buf[1] === 0x49 && buf[2] === 0x2A && buf[3] === 0x0) || (buf[0] === 0x4D && buf[1] === 0x4D && buf[2] === 0x0 && buf[3] === 0x2A)) {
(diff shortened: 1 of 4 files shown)


arthurvr:
Thanks! Needs to be added to the readme as well.

index.js
```diff
@@ -32,6 +32,13 @@ module.exports = function (buf) {
 		};
 	}
 
+	if (((buf[0] === 0x49 && buf[1] === 0x49 && buf[2] === 0x2A && buf[3] === 0x0) || (buf[0] === 0x4D && buf[1] === 0x4D && buf[2] === 0x0 && buf[3] === 0x2A)) && buf[8] === 0x43 && buf[9] === 0x52) {
```
arthurvr: Is there a reason you added this here and not at the bottom? If there is let's add a comment about that.

index.js
```diff
@@ -32,6 +32,13 @@ module.exports = function (buf) {
...
+		return {
+			ext: 'CR2',
+			mime: 'image/x-canon-cr2'
+		}
```
arthurvr: Missing a semicolon

arthurvr:
> The extension is capitalized (.CR2) importing the photos from my camera, so I kept it that way. Should it be in lowercase?

It technically doesn't matter so I'd prefer to be consistent and lowercase it.

index.js
gillstrom: Yeah as I said in the PR, but I'll add one in the code as well.

arthurvr:
Can we get the filesize of the fixture down? 20MB is quite a bunch.

kevva:
> Can we get the size of the fixture down? 20MB is quite a bunch.

CR2 files are pretty large to begin with.

gillstrom:
Fixed everything except the file size. I have troubles finding a smaller one, since all the data is kept in the file even if I change to the size to 1x1.

arthurvr:
I don't like it, but no way around it, so :+1:. Thanks!

arthurvr:
@sindresorhus Can you release this?

sindresorhus:
done

PR #34, opened 2015-09-25
Add dmg support

fasterthanlime:
This adds support to detect 'dmg' files - might be a fringe case for most node.js users, but for [our app](http://github.com/itchio/itchio-app) - it's pretty much something we need to do :)

I'm a bit confused by the [DMG format](https://en.wikipedia.org/wiki/Apple_Disk_Image) to be honest, it seems to exist in many variants.

  * The Wikipedia [List of file signatures](https://en.wikipedia.org/wiki/List_of_file_signatures) says the magic is `78 01 73 0D 62 62 60`
  * A freshly created fixture has `78 01 ED D0 B1 0D 41` instead
  * A file found in the wild has `78 01 73 75 F3 54 08`

The only common denominator is `78 01` and it seems it's not used by any other type supported by `file-type`, so I think it should be fine.

Let me know if anything's wrong with my PR, thanks for the great package!


shinnn:
LGTM

PR #44, opened 2016-01-03
Added ISO Media, MP4 v2 [ISO 14496-14]

yagop:
I created a fixture. A real video file can be found [here](http://techslides.com/sample-webm-ogg-and-mp4-video-files-for-html5).

```bash
$ file fixture/fixture-isomv2.mp4 
fixture/fixture-isomv2.mp4: ISO Media, MP4 v2 [ISO 14496-14]
```


sindresorhus:
Thank you :)

PR #45, opened 2016-01-03
Xo and ava

yagop:
`xo` and `ava` versions in `package.json`. Use `./node_modules/xo/cli.js && ./node_modules/ava/cli.js` for tests


yagop:
Closed, this branch started from `MP4v2-ISOM`

PR #46, opened 2016-01-03
xo and ava versions. Use ./node_modules/xo/cli.js && ./node_modules/a…

yagop:
`xo` and `ava` versions in `package.json`. Use `./node_modules/xo/cli.js && ./node_modules/ava/cli.js` for tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "xo && ava"
+    "test": "./node_modules/xo/cli.js && ./node_modules/ava/cli.js"
   },
   "files": [
     "index.js"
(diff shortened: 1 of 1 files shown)


sindresorhus:
No, there's no point. The current way works fine.

yagop:
Current way works fine but requires `xo` and `ava` installed globally and not many modules uses them.

kevva:
Only Windows needs the absolute path. OS X and Linux resolves to the locally installed one.

sindresorhus:
npm prepends locally installed binaries to the PATH so you can run them by name. No need to globally install.

kevva:
Except, on Windows it doesn't, surprise surprise.

yagop:
Oh it's true, thanks :sweat_smile:

PR #47, opened 2016-01-03
cache node_modules dir

yagop:
Travis cache `node_modules` dir

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ node_js:
   - 'stable'
   - '0.12'
   - '0.10'
+cache:
+  directories:
+    - node_modules


sindresorhus:
No thanks. No point. Travis is fast enough and there are downsides with using a cache.

yagop:
I cant see the downside of using a cached version of `node_modules` for your project. The last Travis job took 59.25s doing the `npm install` which is about 80% (1 min 14 sec).

sindresorhus:
When I run Travis I want to test on a fresh npm install. What if a sub-sub-dependency breaks in a patch version, but it's cached. Then Travis will pass, but will be broken for the user. Not worth the minor speed improvement.

yagop:
I see :+1:

Issue #49, opened 2016-01-20
Do you accept entries that don't have MIME?


T1st3:
I'm thinking of RPM files (which loosely are to Fedora & friends what DEB files are to Debian and Ubuntu). These files can be detected with magic numbers like other file-types, but to my mind, don't have a proper MIME.

sindresorhus:
All entries should have a MIME, but I'm happy to just invent one if you can't even find an unofficial one somewhere on the web. Could be `application/x-rpm`, like `deb`: https://github.com/sindresorhus/file-type/blob/2bcbd763aa26de0430c4cef70cf15f11a6b236c6/index.js#L382

PR #50, opened 2016-01-22
Add `rpm` file type

T1st3:
ref: #49 
Notes:
- Magic number was found at [Gary Kessler's file signatures table](http://www.garykessler.net/library/file_sigs.html).
- The `fixture.rpm` file was created following [Fedora's documentation on creating a RPM package](https://fedoraproject.org/wiki/How_to_create_an_RPM_package)


sindresorhus:
Thanks :)

PR #51, opened 2016-01-26
Add `msi` file type

mastilver:


mastilver:
if you want to know the source of the msi, I took it from there:
https://github.com/dblock/remoteinstall/blob/master/Samples/Sample.msi

readme.md
```diff
@@ -130,6 +130,7 @@ It only needs the first 262 bytes.
 - `rpm`
 - `Z`
 - `lz`
+- `msi`
```
sindresorhus: Can you linkify this to some page about the file type, like wikipedia?

sindresorhus:
Can you also add a `msi` keyword to package.json?

mastilver:
@sindresorhus good catch!
done

mastilver:
@sindresorhus Are you happy with the link?

sindresorhus:
Not really. It doesn't say anything about what the file is.

lolobosse:
Hey there,

I'm very sorry but `0xD0, 0xCF, 0x11, 0xE0, 0xA1, 0xB1, 0x1A, 0xE1` is not not only `MSI`, it's also `doc`, `xls`, `ppt` -> https://en.wikipedia.org/wiki/List_of_file_signatures

Try to parse a `doc` you'll see... 😢 

https://github.com/lolobosse/file-type/tree/PR55-bug (yeah I thought it was 55 and not 51, I'm retarded...)

lolobosse:
Btw, I'm not the only one which referenced the problem (libmagic has it too)

https://github.com/threatstack/libmagic/blob/1249b5cd02c3b6fb9b917d16c76bc76c862932b6/magic/Magdir/msdos#L898

mastilver:
@lolobosse Yeah, you're right... http://www.garykessler.net/library/file_sigs.html :(

So there is no way to tell them apart? Even using bits after the header?

karlhiramoto:
@mastilver yes you can tell them apart, however you need to parse a lot more data.
olefile is a python library that does this.  see https://github.com/decalage2/olefile

lolobosse:
@karlhiramoto Yeah and then we get out of the scope of the lib, no? 

What do you think @sindresorhus? 

It's actually related to my #112 issue

PR #53, opened 2016-02-05
feat: add opus file type

t1st3:
PR adds [opus](https://en.wikipedia.org/wiki/Opus_(audio_format)) file type

While opus is an ogg-like file, its detection has to be before the ogg detection.

Relevant:
[opus spec @ xiph](https://wiki.xiph.org/OggOpus), which details file header and signature
[RFC draft for opus](https://tools.ietf.org/html/draft-ietf-codec-oggopus-10)

Fixture was found at [opus examples @ xiph](https://people.xiph.org/~giles/2012/opus/)


index.js
```diff
@@ -218,6 +218,14 @@ module.exports = function (buf) {
 		};
 	}
 
+	// needs to be before `ogg` check
+	if (buf[28] === 0x4F && buf[29] === 0x70 && buf[30] === 0x75 && buf[31] === 0x73 && buf[32] === 0x48 && buf[33] === 0x65 && buf[34] === 0x61 && buf[35] === 0x64) {
+		return {
+			ext: 'opus',
+			mime: 'audio/ogg; codecs=opus'
```
sindresorhus: According to this it's `audio/opus`: https://tools.ietf.org/html/rfc7587#section-6

t1st3: You're right, `audio/opus` is clearer and makes more sense. All the 3 first links given above mention the other mime, which is kind of misleading... Anyway, I just updated that MIME.

sindresorhus:
Thanks :)

PR #57, opened 2016-03-19
Add `docx`, `pptx` and `xlsx` file types

overra:
Initially used `file-type` to get the file type of a `docx` buffer but it was returning `'zip'`. This is because OOXML files are just a zipped file structure with different file extensions. The whole file needs to be read in order to unzip and parse the necessary file to determine what type of OOMXL file type it is. Because of this, I removed the `read-chunk` dependency during and testing I just pass the full file buffer to `fileType()`.

I learned about the file type signature on Gary Kessler's [File Signature Table](http://www.garykessler.net/library/file_sigs.html) and found the mime type [here](http://www.freeformatter.com/mime-types-list.html).


sindresorhus:
One thing I like with this module is that you don't have to read in the whole file to detect the file type. Adding these would remove that benefit for very limited gain (I've personally never have had to detect these file types). I suggest you create a higher-level module that depends on this one, but adds those file types.

throrin19:
This feature is really helpful for our project besause we can't check the correct mime-type with the upload object :

In windows the mime-type is equal to `application/octet-stream` for .docx et .xlsx
In linux, the mime-type is different.

In this condition, it's impossible to check correctly the format. Please make this merge request

Issue #67, opened 2016-11-22
MKV detected as WebM


mifi:
It seems that the check in this library is too simple, because the strings "matroska" and "webm" are not always in a fixed position.

libmagic (used by the unix "file" utility) detects it correctly. Looking at its source, it does a search for the string "webm" and "matroska" preceded by \x42\x48 and a wildcard byte, in the first 4096 bytes of the file.

See source: https://github.com/threatstack/libmagic/blob/master/magic/Magdir/matroska

This would however require more complex checking logic. And I don't know if it's the scope of this project to be as exact as possible in detection.

At the very least I think that the checking code should be changed so that it FIRST tries to match the "webm" magic string, and if it fails to match that it should fallback to matroska (given that the first 4 bytes match of course.) The way it works now, it falls back to WebM, which is a very limited subset of matroska.

See attached mkv file:
[test.mkv.zip](https://github.com/sindresorhus/file-type/files/606313/test.mkv.zip)

sindresorhus:
I want this module to be as exact as possible. I have no problem adding extra logic to achieve that. 

PR welcome :)

mifi:
Cool! Are you also open for upping the required node version to, say 4? Because working with buffers in 0.10 is not very pleasant, I think

mifi:
The holy grail would be to build JS a parser for the magic database and bundle it in the module, so we don't have to reimplement all the formats manually in code... But that's no trivial task unfortunately.

sindresorhus:
> Are you also open for upping the required node version to, say 4? Because working with buffers in 0.10 is not very pleasant, I think

Done: https://github.com/sindresorhus/file-type/commit/2bea1a2b3e701c302fea92cf269b01ad0101b2ff

> The holy grail would be to build JS a parser for the magic database and bundle it in the module, so we don't have to reimplement all the formats manually in code...

Oh yeah. That would be so much better. Although this module already covers most of the common formats. Mind opening an issue about that? Maybe someone someday will see it and take the challenge.

PR #69, opened 2016-11-25
Improve MKV/WebM detection

mifi:
Inspired by:
https://github.com/threatstack/libmagic/blob/master/magic/Magdir/matroska

See #67 
Also I'm not sure whether we will need more than the first 262 bytes to detect all mkv. (libmagic uses the first 4096 (+4?)). However it seems to be working so I guess it's fine for now


index.js
```diff
@@ -168,19 +168,28 @@ module.exports = function (buf) {
...
-		};
-	}
-
+	// https://github.com/threatstack/libmagic/blob/master/magic/Magdir/matroska
 	if (buf[0] === 0x1A && buf[1] === 0x45 && buf[2] === 0xDF && buf[3] === 0xA3) {
-		return {
-			ext: 'webm',
-			mime: 'video/webm'
-		};
+		const sliced = buf.subarray(4, 4 + 4096);
```
sindresorhus: This is a TypedArray method, not a Buffer one. While Buffers inherit from TypedArray, they're not entirely the same and depending on something like this feels unsafe. I would just use `buf.slice()`, which actually does the same thing as `TypedArray#subarray()`.

sindresorhus: You need to update the readme about how many bytes the user needs to read in.

index.js
```diff
@@ -168,19 +168,28 @@ module.exports = function (buf) {
...
+		const idPos = sliced.findIndex((e, i, a) => a[i] === 0x42 && a[i + 1] === 0x82);
```
sindresorhus: This is also not a Buffer method.

sindresorhus: `(e, i, a)` => `(el, i, arr)`

index.js
mifi: The reason why I used TypedArray methods is because this project can be used from the browser too, so I wanted to not tie it to node.js. Also, subarray creates a view over the underlying data, while slice copies the bytes, which is slower and should be unnecessary.

Regarding the number of bytes the user needs to read, I thought it would be OK to use the same amount as we already do, because it seems to work on the fixtures and all the MKV/WebM files that I tested, and instead we can increase it if someone encounters a file that happens to fail. However if you want, sure I can increase it.

index.js
mifi: Since Buffer inherits from TypedArray I thought it would be ok to use this browser-compatible function. Otherwise we would have to detect if he obj is a Buffer or TypedArray, and then do it differently?

index.js
sindresorhus: > The reason why I used TypedArray methods is because this project can be used from the browser too, so I wanted to not tie it to node.js

Yes, but only with Browserify, which already shims the Buffer parts.

> Also, subarray creates a view over the underlying data, while slice copies the bytes, which is slower and should be unnecessary.

Not according to the docs:

> while ArrayBuffer#slice() creates a copy of the slice, the implementation of Buffer#slice() creates a view over the existing Buffer without copying - https://nodejs.org/api/buffer.html#buffer_buffers_and_typedarray

mifi: I wasnt aware that browesrify does this. I guess if webpack also does this, then it's not a big problem.

Regarding slice, I was looking at the doc for TypedArray, which is what is used in the readme browser example of this project. It says that slice will do a copy.

I agree that we should read enough bytes.

sindresorhus: > which is what is used in the readme browser example of this project

Ah, forgot I had an example with that. I would prefer to work in one format though, so we should either convert the input to Buffer or the other way around at the top. Up to you. Whatever the easiest for you.

For reference, from TypedArray to Buffer:

```js
const buffer = new Buffer(new Uint8Array(x));
```

mifi: Ok. IMO we should wrap in Uint8Array since it doesn't require shims in the browser like Buffer does.

sindresorhus: Let's go with `Uint8Array`.

sindresorhus:
Woot! Thank you @mifi. Glad to see the detection improved.

sindresorhus:
I've invited you as a collaborator on this repo. No worries if you're not interested :)

mifi:
Gr8 thanks 🏄

Issue #70, opened 2016-12-22
Arrow functions are not compatible with UglifyJS or IE11


cveilleux:
The latest 4.0.0 release uses arrow functions but that is not quite ready for prime time.

http://caniuse.com/#feat=arrow-functions

Also, UglifyJS does not support arrow functions.

We bundle our javascript with Webpack and use UglifyJS to minify our final bundle. The 4.0 release breaks because of the arrow functions.

Progress is good and tolling support for arrow function will get more and more widespread, but I suggest may be adding a warning on the 4.0 release / branch that some users might prefer to stay on the 3.x branch?

sindresorhus:
Nah, the main target is Node.js. Up to you to transpile if you want it in the browser. Relevant discussion: https://github.com/facebookincubator/create-react-app/issues/1125#issuecomment-264217076

PR #71, opened 2017-01-04
Babelify ES6 code to ES5 for non-babel-y consumers

kleingtm:


sindresorhus:
Duplicate of https://github.com/sindresorhus/file-type/issues/70.

kleingtm:
When babel stops shipping with its default as ignoring node_modules, then the argument from the discussions in #70 hold. Until then, I personally believe the baseline cannot possibly be ES6. I also think the burden on the consumer to regex what does and does not need to be transpiled is much greater than the burden of a module maintainer to simply add a compile step. Just my two cents. That said, it's your module and you can, of course, do what you like with it 😄

philipstanislaus:
In need of this feature, I forked and published changes similar to this PR to npm: 

https://www.npmjs.com/package/file-type-es5

The `file-type-es5` repository includes an ES5 build. I will keep the repository in sync.

Thanks to @sindresorhus for this awesome library!

PR #72, opened 2017-01-09
Add JSON detection

LinusU:
Adds json detection by looking for a file that either

1) begins with `{` and has whitespace or `"` as the next character, or
2) begins with `[` and has whitespace or beginning of any json value as next character

Whitespace is defined by the json standard as:

```text
  ws = *(
            %x20 /              ; Space
            %x09 /              ; Horizontal tab
            %x0A /              ; Line feed or New line
            %x0D                ; Carriage return
        )
```


sindresorhus:
Not sure if this is a good idea. `{"` could be anything.

@mifi Thoughts?

mifi:
I agree that {" could be anything. and json isn't really a file format. Also json could start with a whitespace or newline, right?
Also unix "file" utility doesn't recognize json. I don't think it's a good idea either.

qzb:
Perfectly valid JSON can also start with `"`, `n`, `t`, `f`, or a digit.

LinusU:
Very good points from all of you 👍 I think that the correct course of action is to close this

PR #80, opened 2017-04-14
Add WebAssembly support

wanderer:
this add support for detecting webassembly binarys


package.json
```diff
@@ -102,6 +102,10 @@
     "xo": "*"
   },
   "xo": {
-    "esnext": true
+    "esnext": true,
+    "rules": {
```
SamVerschueren: Do not turn these rules off, I will fix them in master later today.

SamVerschueren: Fixed them in master. So now you can rebase upon master and it should work.

sindresorhus:
You also need to add a link in the readme and package.json keyword.

Can you link to the source for the "magic number" used here?

wanderer:
ok @sindresorhus documentation on the magic number and readme updated!

sindresorhus:
I meant here, not in the source code, but I can fix that. For reference:

```js
// The Magic number is  0x6d736100 (i.e., '\0asm') see https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#high-level-structure
```

sindresorhus:
Thank you :)

Issue #77, opened 2017-03-20
Blender file support


webbrandon:
I have a project that will now need to process `blend` files.  It would be wonderful to not have a customization that needs to figure this exception out.

Ref: http://www.atmind.nl/blender/mystery_ot_blend.html

sindresorhus:
Sure. Seems like it's easy to add. Just need to detect `BLENDER` (in bytes) at the start of the file. Pull request welcome.

webbrandon:
thank you!  I will try and get to add this when I have a moment.

kevva:
Fixed in https://github.com/sindresorhus/file-type/commit/07571ddcf7d8c3bad4c5530af8c9d0451ae9e09d.

PR #84, opened 2017-06-03
Add Blender file format (#77)

thebravyone:


mifi:
I don't think package.lock should be included here?

SamVerschueren:
Nope, it shouldn't

kevva:
Thanks 👍

PR #85, opened 2017-06-05
Update `font` media types

alrra:
The `font` top-level media type has been registered:

* https://tools.ietf.org/html/rfc8081
* https://www.iana.org/assignments/media-types/media-types.xhtml#font


sindresorhus:
I guess this would be considered a breaking change?

SamVerschueren:
Yes, breaking.

PR #86, opened 2017-06-06
Adding support for MPEG transport stream

set-killer:
More info:

https://en.wikipedia.org/wiki/MPEG_transport_stream
http://www.garykessler.net/library/file_sigs.html (see the .ts file format)

### DIFF
--- a/index.js
+++ b/index.js
@@ -534,6 +534,13 @@ module.exports = input => {
 		};
 	}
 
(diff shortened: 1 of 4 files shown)


kevva:
Needs test.

set-killer:
Updated.

kevva:
Thanks :). Not much to go on so let's hope we don't hit a lot of false positives here.

PR #98, opened 2017-08-08
Add support for yet another MP4 variant

karlhiramoto:
Add a check to  detect mp4 video file from http://fsi.stanford.edu/sites/default/files/video_2.mp4

I didn't add a test as this video file is rather large and I'm not sure about license.


mifi:
You can truncate the file to only the first needed bytes. Also, can you reference any documentation or code from libmagic for what kind of file/variant this is, and why this special check is needed?

karlhiramoto:
OK added a test fixture. 

you can find this format listed as "mp42"   at http://www.garykessler.net/library/file_sigs.html

Libmagic detects it as "ISO Media, MP4 v2 [ISO 14496-14]"

mifi:
I cannot find the byte sequence you committed on the page you linked

karlhiramoto:
@mifi  Do you care if we call  'm4v' a 'mp4'?   You could argue that m4v is a variant of mp4.   Do you prefer to generalize or have more detailed format info?  In my use case I'm more interested in  knowing if its a video, audio, or image file.  I don't care too much about the fine details. 

If your OK calling m4v  a mp4  the code can be simplified.  In this case  a easier match would be 
```check([0x0, 0x0, 0x0]) && check([0x66, 0x74, 0x79, 0x70], {offset: 4})``` 

If you want more detail I think we want something like:
```check([0x0, 0x0, 0x0]) && check([0x66, 0x74, 0x79, 0x70, 0x34, 0x32], {offset: 4})```

I've read at http://www.file-recovery.com/m4a-signature-format.htm  That this XX  here is the length of the header in big endian  `00 00 00 XX 66 74 79 70 `

Another reference https://github.com/PidgeyL/Forensic-File-Handler
Have a look at the string  `[4 bytes] 66 74 79 70 `

Another reference:
http://www.asecuritysite.com/information/mp4?file=http%3A%2F%2Fbillatnapier.com%2Fnetworksims_wlc03.mp4

**Comment by mifi** [comment]
> @mifi Do you care if we call 'm4v' a 'mp4'? You could argue that m4v is a variant of mp4. Do you prefer to generalize or have more detailed format info? In my use case I'm more interested in knowing if its a video, audio, or image file. I don't care too much about the fine details.

That's a good question. Currently it seems like we are detecting some files as [m4v](https://github.com/sindresorhus/file-type/blob/master/index.js#L202) and some [mp4](https://github.com/sindresorhus/file-type/blob/master/index.js#L195).

[libmagic](https://github.com/threatstack/libmagic/blob/1249b5cd02c3b6fb9b917d16c76bc76c862932b6/magic/Magdir/animation) classifies all these video types as video/mp4, so i think maybe we should do that too, but that would be a breaking chnage.


> If your OK calling m4v a mp4 the code can be simplified. In this case a easier match would be
> check([0x0, 0x0, 0x0]) && check([0x66, 0x74, 0x79, 0x70], {offset: 4})

That would match everything (even audio and 3gpp) as video/mp4, wouldn't it?


> I've read at http://www.file-recovery.com/m4a-signature-format.htm That this XX here is the length of the header in big endian 00 00 00 XX 66 74 79 70

Yes it seems that libmagic ignores the first 4 bytes including this length.

Analyzing the code a bit it seems like most checks under [video/mp4](https://github.com/sindresorhus/file-type/blob/master/index.js#L180) boil down to the same basic check:

## video/mp4

**ftyp**
0x0, 0x0, 0x0, 0x18, 0x66, 0x74, 0x79, 0x70

**ftyp**
0x0, 0x0, 0x0, 0x20, 0x66, 0x74, 0x79, 0x70

**3gp5**
0x33, 0x67, 0x70, 0x35

**ftypmp42 ...**
0x0, 0x0, 0x0, 0x1C, 0x66, 0x74, 0x79, 0x70, 0x6D, 0x70, 0x34, 0x32 .... 0x6D, 0x70, 0x34, 0x31, 0x6D, 0x70, 0x34, 0x32, 0x69, 0x73, 0x6F, 0x6D

**ftypisom**
0x0, 0x0, 0x0, 0x1C, 0x66, 0x74, 0x79, 0x70, 0x69, 0x73, 0x6F, 0x6D

**ftypmp42 \0\0\0**
0x0, 0x0, 0x0, 0x1C, 0x66, 0x74, 0x79, 0x70, 0x6D, 0x70, 0x34, 0x32, 0x0, 0x0, 0x0, 0x0

## video/x-m4v

**ftypM4V**
0x0, 0x0, 0x0, 0x1C, 0x66, 0x74, 0x79, 0x70, 0x4D, 0x34, 0x56

# Conclusion
I think the whole `video/mp4` block could be simplified down to:

```
check([0x33, 0x67, 0x70, 0x35]) || // 3gp5
(
  check([0x0, 0x0, 0x0]) && check([0x66, 0x74, 0x79, 0x70], {offset: 4}) && (
    check([0x6D, 0x70, 0x34, 0x31], {offset: 8}) || // mp41
    check([0x6D, 0x70, 0x34, 0x32], {offset: 8}) || // mp42
    check([0x69, 0x73, 0x6F, 0x6D], {offset: 8}) || // isom
    check([0x69, 0x73, 0x6F, 0x32], {offset: 8}) || // iso2
    check([0x6D, 0x6D, 0x70, 0x34], {offset: 8}) || // mmp4
  )
)
```

Also video/x-m4v check could be merged into this new video/mp4 check but it would be breaking.
```
  check([0x4D, 0x34, 0x56], {offset: 8}) || // M4V
```

mifi:
@sindresorhus do you agree with this simplification in order to make it similar to libmagic?
I don't think it should be too breaking of a change because we are covering the old cases too.

sindresorhus:
> do you agree with this simplification in order to make it similar to libmagic?

👍 

> Also do you think we should also identify m4v as video/mp4 like libmagic, instead of video/x-m4v which we do now. (requires major semver bump)

Yes

PR #100, opened 2017-08-27
Improve performance

magicode:


index.js
```diff
@@ -5,9 +5,9 @@ const oxmlContentTypes = toBytes('[Content_Types].xml');
 const oxmlRels = toBytes('_rels/.rels');
 
 module.exports = input => {
-	const buf = new Uint8Array(input);
```
sindresorhus: This is intentional, for browser support.

sindresorhus:
I appreciate the PR, but it's missing a benchmark proving the numbers you claim :)

index.js
magicode: this make new buffer it It takes performance.
Why an array of numbers is not good for checking?

magicode:
try this

```javascript

const fs = require('fs');
const filetype = require('.');

var file = fs.readFileSync(__dirname + "/fixture/fixture.png");

var time = 3000;
var starttime = Date.now();
var endtime = starttime + time;
var count = 0;
for(;Date.now()<endtime;count++){
    filetype(file);
}

console.log("%d ops/s" ,~~(count/((Date.now()-starttime)/1000)));
```

sindresorhus:
@mifi Any thoughts on this?

index.js
```diff
@@ -5,9 +5,9 @@ const oxmlContentTypes = toBytes('[Content_Types].xml');
...
+	const buf = input;
 
-	if (!(buf && buf.length > 1)) {
```
mifi: It's a bit unclear why u compare with -1, why not use buf[1] !== undefined ?

magicode: It checks that it is also a number

mifi:
I dunno if some of our users are sending in an array, but if they do, some file type checks will fail when removing the wrapping, like those using `buf.subarray()`, because array does not have this method.
Our docs however say clearly that we input should be Buffer or Uint8array. Still if we do this change i think we should bump major version.

Also, is it the removal of the wrapping or the length check that boosts performance?

index.js
mifi: Why is that needed? also `'' > -1` is `true` which is obviously not a number

magicode: It's a check that on the way.
It's not important

magicode:
What's important for performance.
It does not wrap if it Buffer or Uint8array.

mifi:
Ok, I think then you can revert the `buf[1] > -1` change, if it is not important for performance.
Also, don't assign `const buf = input;`, instead rename argument `input` to `buf`.
Then I think this can be merged if @sindresorhus agrees.

sindresorhus:
Ok, sure, when @mifi's feedback is addressed.

PR #101, opened 2017-09-02
Check for MP3 header at various starting offsets

### LINKED ISSUE #75: Unrecognized mp3 file
I have some mp3 file not recognized by file-type:

```bash
❯ file /usr/local/Wowza/content/20151027_Consiglio_01.mp3
/usr/local/Wowza/content/20151027_Consiglio_01.mp3: MPEG ADTS, layer III,  v2.5,  64 kbps, 11.025 kHz, Monaural
```

first bytes in file are 0xffe380, while file-type check for 0x494433 or else 0xfffb.

https://en.wikipedia.org/wiki/List_of_file_signatures only list the two checks actually done by file-type, but the magic header file used by posix `file` command seems to list a more complex check:

```
# MP3, M1A
0       beshort&0xFFFE  0xFFFA         MPEG ADTS, layer III, v1
# rates
>2      byte&0xF0       0x10           \b,  32 kBits
>2      byte&0xF0       0x20           \b,  40 kBits
>2      byte&0xF0       0x30           \b,  48 kBits
>2      byte&0xF0       0x40           \b,  56 kBits
>2      byte&0xF0       0x50           \b,  64 kBits
>2      byte&0xF0       0x60           \b,  80 kBits
>2      byte&0xF0       0x70           \b,  96 kBits
>2      byte&0xF0       0x80           \b, 112 kBits
>2      byte&0xF0       0x90           \b, 128 kBits
>2      byte&0xF0       0xA0           \b, 160 kBits
>2      byte&0xF0       0xB0           \b, 192 kBits
>2      byte&0xF0       0xC0           \b, 224 kBits
>2      byte&0xF0       0xD0           \b, 256 kBits
>2      byte&0xF0       0xE0           \b, 320 kBits
# timing
>2      byte&0x0C       0x00           \b, 44.1 kHz
>2      byte&0x0C       0x04           \b, 48 kHz
>2      byte&0x0C       0x08           \b, 32 kHz
# channels/options
>3      byte&0xC0       0x00           \b, Stereo
>3      byte&0xC0       0x40           \b, JntStereo
>3      byte&0xC0       0x80           \b, 2x Monaural
>3      byte&0xC0       0xC0           \b, Monaural
#>1     byte            ^0x01          \b, Data Verify
#>2     byte            &0x02          \b, Packet Pad
#>2     byte            &0x01          \b, Custom Flag
#>3     byte            &0x08          \b, Copyrighted
#>3     byte            &0x04          \b, Original Source
#>3     byte&0x03       1              \b, NR: 50/15 ms
#>3     byte&0x03       3              \b, NR: CCIT J.17
```

karlhiramoto:
I found some mp3 content online that did not validate.  I added the first 128 bytes of them to test fixtures, the files I found that did not validate are:
 * https://www.bensound.com/royalty-free-music?download=scifi
 * https://www.bensound.com/royalty-free-music?download=sunny

Doing some investigation it seems MP3 headers can occur at any point in the file, since MPEG is intended to be used for broadcast.  In practice It seems that most players do check for the headers at various locations.   In this patch I'm just looking over the first 2 bytes. 

For example mp3check can look for headers throughout the file:
https://github.com/pjjw/mp3check/blob/master/src/mp3check.cc#L460

Other references on MP3 file format:
* http://www.datavoyage.com/mpgscript/mpeghdr.htm
* http://www.mp3-converter.com/mp3codec/mp3_anatomy.htm

---

Fixes #75


index.js
```diff
@@ -17,7 +17,11 @@ module.exports = input => {
 		}, opts);
 
 		for (let i = 0; i < header.length; i++) {
-			if (header[i] !== buf[i + opts.offset]) {
+			if (opts.mask) {
```
sindresorhus: Can you add a comment about what the `mask` option does? For posterity.

sindresorhus:
Do all of these `mp3` files identify correctly with the `file` command?

sindresorhus:
Might be worth looking into how https://github.com/threatstack/libmagic detect MP3.

karlhiramoto:
@sindresorhus   added the comment where requested.     ```file``` aka libmagic  does NOT correctly detect these examples.  They do work though with mp3check,  ffmpeg,  mpg321, vlc, windows and macOS default players.     In my experience libmagic is not the greatest, and should not be a model of how to do things.   I've removed libmagic from other projects I worked on.    My biggest libmagic complaints are:
  *   sometimes incorrectly identifies files as wrong type
  *   sometimes does not identify them at all or marks them as 'data'
  *   It's not thread-safe, for multi threaded applications.

mifi:
Isn't it safer to use the mask 0b11100**11**0 instead of 0b11100000 (0xe0)? And match against 0b11100**01**0.
That way we match that it is a mpeg layer III file (Sign C under http://www.datavoyage.com/mpgscript/mpeghdr.htm)
Else we would also match non layer 3 files.

karlhiramoto:
@sindresorhus agreed fe8faa2  will do 0b11100010 (0xE2) so will will match Layer 3.

index.js
karlhiramoto: comment in c433cf3b6b23691709a3f7c4fa2114a74b190ccb

sindresorhus:
Thanks for looking into and fixing this @karlhiramoto :)

Issue #29, opened 2015-08-06
OGV file identified as audio/ogg


brunobg:
Tried to identify an OGV file and it was interpreted as audio:

{ ext: 'ogg', mime: 'audio/ogg' }

sindresorhus:
The OGG container has the same magic number (marker to identify the type) for both audio and video. So not sure whether there's any way to tell them apart.

https://en.wikipedia.org/wiki/Ogg

brunobg:
Apparently you have to read the ogg container data. It certainly is possible. `file` does it: 

```$ file video.ogv
video.ogv: Ogg data, Theora video```

If you can't, I suggest you return `application/ogg` instead of `audio/ogg` at least, and keep the original file extension instead of converting it to ".ogg".

Alcotana:
I encountered this problem too.

PR #104, opened 2017-09-24
OGG: Check contents of OGG container which can be audio or video

karlhiramoto:
This pull request will fix detection of OGG files as audio, when there are video types of OGG

Libmagic does checks around byte 28 for content within OGG container
https://github.com/threatstack/libmagic/blob/master/magic/Magdir/vorbis

Fixes https://github.com/sindresorhus/file-type/issues/29


karlhiramoto:
@mifi  I  defaulted to keep backwards compatibility keep old behavior with unknown file. 

I agree though it's better to follow specs:
  * https://www.iana.org/assignments/media-types/application/ogg
  * https://tools.ietf.org/html/rfc5334

This PR should be a major version change as it might break anyone's code that was expecting .ogg  audio/ogg.

There are other OGG types, but I included tests and fixtures for the ones I could easily find samples of.

mifi:
Yes I agree that it should be a major bump. Other file types can be added later. Thanks @karlhiramoto

PR #106, opened 2017-10-04
Add 3gp

medez:


test.js
```diff
@@ -90,7 +91,8 @@ const names = {
 	lz: ['fixture.tar'],
 	Z: ['fixture.tar'],
 	mkv: ['fixture', 'fixture2'],
-	mxf: ['fixture']
+	mxf: ['fixture'],
+	'3gp': ['fixture']
```
kevva: This isn't needed since it'll test `fixture.3gp` by default. These are only needed for fixtures that needs special handling.

kevva:
Thanks :)

PR #108, opened 2017-10-10
Add support for JPEG 2000 

bjornstar:
`image/jp2`
`image/jpm`
`image/jpx`
`video/mj2`


bjornstar:
I saw larger files so I thought it was preferred to have full files instead of just the headers.

I have resized the jp2 image and just truncated the others.

sindresorhus:
Thanks :)

PR #109, opened 2017-10-15
Improve performance

magicode:
fix PR #100
Improve performance up 30 times 
this benchmark 
```js
 
const fs = require('fs');
const filetype = require('.');

var file = fs.readFileSync(__dirname + "/fixture/fixture.png");

var time = 3000;
var starttime = Date.now();
var endtime = starttime + time;
var count = 0;
for(;Date.now()<endtime;count++){
    filetype(file);
}

console.log("%d ops/s" ,~~(count/((Date.now()-starttime)/1000)));
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,7 @@ const oxmlContentTypes = toBytes('[Content_Types].xml');
 const oxmlRels = toBytes('_rels/.rels');
 
 module.exports = input => {
-	const buf = new Uint8Array(input);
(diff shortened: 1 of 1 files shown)


mifi:
Thanks @magicode. Good idea to check the type, but there are 2 issues:
1. Buffer is already instanceof Uint8array, so no need to check for Buffer
2. Buffer might not exist in a browser environment so that check might give a ReferenceError

PR #113, opened 2017-11-10
Add XML file format

StomCarlo:


mifi:
See previous PRs for instructions

package.json
```diff
@@ -40,7 +40,8 @@
     "uint8array",
     "jpg",
     "png",
-    "gif",
+		"gif",
```
mifi: seems indentation changed here

index.js
```diff
@@ -52,6 +52,13 @@ module.exports = input => {
 		};
 	}
 
+	if (check([0x3C, 0x3F, 0x78, 0x6D, 0x6C, 0x20])) {
```
sindresorhus: What if there's an empty line before `<?xml version="1.0" encoding="UTF-8"?>`?

StomCarlo: It does not work, but the standard specify that an xml document has to start with the xml declaration ( https://www.w3.org/TR/xml/#sec-prolog-dtd ) and new lines are converted in a special character (https://www.w3.org/TR/xml/#sec-line-ends), so it is like if the document is not beginning with the xml declaration.

package.json
StomCarlo: I fixed this in the last push

index.js
sindresorhus: Can you use the `toBytes()` method here for readability reasons?

https://github.com/sindresorhus/file-type/blob/4408621b5b8ea17911306be696c6fa1201bce34a/index.js#L2

StomCarlo: I tried to use it, but it does not return the correct magic number, the returned one is correct except for the last element, which is 0X2F instead of 0X20, if I slice the returned magic buffer ignoring the last element it works, but i'm not sure that this will always work fine.

mifi: I don't understand why that shouldn't work
```
Array.from('<?xml ').map(c => c.charCodeAt(0))
(6) [60, 63, 120, 109, 108, 32]

[0x3C, 0x3F, 0x78, 0x6D, 0x6C, 0x20]
(6) [60, 63, 120, 109, 108, 32]
```

StomCarlo: well sorry I was using the wrong offset

PR #115, opened 2017-11-24
Adding Support for ReactNative for Ease of Use

prscX:
Hi @sindresorhus,

First of all I would like to appreciate for creating such a nice library. 

I have created a [ReactNative Wrapper](https://github.com/prscX/react-native-file-type) for the ease of use. Please find below steps we are performing in order to retrieve file meta data:

- It accepts file path stored in device local storage (iOS, Android, Windows) 
- Internally we are using [RNFS](https://github.com/itinance/react-native-fs/) to fetch the file
- Once file is retrieved we are using [js-base64](https://github.com/dankogai/js-base64) to generate UTF8Array Buffer
- Then we are passing generated buffer to [file-type](https://github.com/sindresorhus/file-type) in order to get the meta data

I have added the reference of this library in [file-type](https://github.com/sindresorhus/file-type) README, so that in case any one looking to use this library can easily use the same

Can you please merge the raised PR. Please let me know in case any discussion or changes are requried

Thanks
Pranav

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -173,6 +173,11 @@ It only needs the first 4100 bytes.
 - [Sindre Sorhus](https://github.com/sindresorhus)
 - [Mikael Finstad](https://github.com/mifi)
 
+## Third Party Bindings
+
+### React Native
+Thanks to @prscX, you may now use this library with [React Native](https://github.com/facebook/react-native) via the module [here](https://github.com/prscX/react-native-file-type)
+
 
 ## License
 


sindresorhus:
I'm not familiar with React Native, but going from binary to base64 and then to binary again sounds extremely inefficient. Seems like you could just this https://github.com/wkh237/react-native-fetch-blob and be able to use `file-type` directly.

prscX:
Thanks @sindresorhus for reviewing the request. I will go through the [react-native-fetch-blob](https://github.com/wkh237/react-native-fetch-blob)

PR #117, opened 2017-11-30
Add Mobipocket file format

sbugert:
This adds support for mobi files.

Is there a specific order in which checks are added to index.js or is it okay to append it to the end?


sindresorhus:
At the end is fine.

Can you also add the `mobi` keyword to package.json?

Can you link to where you found the file signature / magic number for `mobi`?

readme.md
```diff
@@ -116,6 +116,7 @@ It only needs the first 4100 bytes.
 - [`amr`](https://en.wikipedia.org/wiki/Adaptive_Multi-Rate_audio_codec)
 - [`pdf`](https://en.wikipedia.org/wiki/Portable_Document_Format)
 - [`epub`](https://en.wikipedia.org/wiki/EPUB)
+- [`mobi`](https://wiki.mobileread.com/wiki/MOBI#Format) - Mobipocket
```
sindresorhus: I should have been clearer. I just meant link it as a comment here. This should still point to the Wikipedia article.

sbugert: Ah, I see! I should have seen all the other Wikipedia links.

sbugert:
https://wiki.mobileread.com/wiki/MOBI#Format
So Mobipocket is an extension of [PalmDOC](https://wiki.mobileread.com/wiki/PDB#Palm_Database_Format) with Creator ID of MOBI and a Type of BOOK from offset 60 which is exactly what is checked.

PR #118, opened 2017-12-01
Add support for a variant of mpg videos

drimer:
I've added a sample, which is a perfectly valid video. The headers are
000001b3, which is slightly different from the only one that was being
supported.


sindresorhus:
Can you link to some docs/source about this?

drimer:
https://en.wikipedia.org/wiki/List_of_file_signatures

The last column to the right has the two possible values. If you'd like, I'm happy to amend the commit message an add this to it.

sindresorhus:
`fixture-mpg.mpg` => `fixture2.mpg`

sindresorhus:
You need to include the extra fixture in the test, like done here: https://github.com/sindresorhus/file-type/blob/4ad8b6e8cd8feab2f827f942d77d34f74e1b5d85/test.js#L105

drimer:
Oh, yeah. I thought it'd be automatically picked up by the tests. I've just amended that. Also added the existing `fixture.mpg` to the tests, as it was missing.

PR #125, opened 2018-02-12
remove 1 bit from Mp3  signature  to match additional files

karlhiramoto:
changing one bit related to #101
The bit removed in the check corresponds to the "Layer description"  which distinguishes between MPEG   Layer 1,2, or 3  The example in 11e7ed2  is  layer 2 audio


sindresorhus:
MP2 is a different file format and should not be matched as MP3: https://en.m.wikipedia.org/wiki/MPEG-1_Audio_Layer_II

karlhiramoto:
closing.   This pull request is superseded by  #127

PR #124, opened 2018-02-01
Make 3gp signature more lenient to match spec

dominic-p:
This is a small tweak to the 3gp detection signature to bring it in line with the spec [found on Wikipedia](https://en.wikipedia.org/wiki/List_of_file_signatures). The original signature missed at least one valid .3gp file I encountered. Feedback is welcome.

Fixes #105.

PR #126, opened 2018-02-12
Make 3gp signature more lenient to match spec

### LINKED ISSUE #105: Support for 3gp
Hello, 

Could you add the support of 3gp files ? For information, 3gp files are identified by the string "ftyp3gp4" at the offset 3.

Best regards,
Seb

BrandonSmith:
This pull request adds a fixture for pull request (#124) @dominic-p provided that fixes #105.


test.js
```diff
@@ -93,6 +93,7 @@ const types = [
 ];
 
 const names = {
+	'3gp': ['fixture', 'fixture2'],
```
karlhiramoto: should be consistent and not use quotes.

BrandonSmith: @karlhiramoto I'm open to suggestions. However, unquoting `'3gp'` would be a syntax error.

sindresorhus:
Same comment as https://github.com/sindresorhus/file-type/pull/127#issuecomment-365172265

test.js
sindresorhus: Quoted is fine here as it's needed.

dominic-p:
The readme and package.json already list 3gp support. All this PR does is fix the signature. Unless I'm missing something...

sindresorhus:
@dominic-p You're right. My mistake.

Issue #123, opened 2018-01-23
Add support for wma and wmv files


Frezzaldo:
:)

Borewit:
Distinguish between WMA and WMV formats blocks Borewit/music-metadata#110

PR #138, opened 2018-05-07
Distinguish between WMA and WMV formats

karlhiramoto:
Closes https://github.com/sindresorhus/file-type/issues/123

### DIFF
--- a/index.js
+++ b/index.js
@@ -317,7 +317,7 @@ module.exports = input => {
 		};
 	}
 
-	// ASF_Header_Object is first 64 bits
+	// ASF_Header_Object first 80 bytes
 	if (check([0x30, 0x26, 0xB2, 0x75, 0x8E, 0x66, 0xCF, 0x11, 0xA6, 0xD9])) {
 		// Search for header should be in first 1KB of file.
 
--- a/index.js
+++ b/index.js
@@ -317,7 +317,31 @@ module.exports = input => {
 		};
 	}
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -317,7 +317,31 @@ module.exports = input => {
...
+			if (check([0x40, 0x9E, 0x69, 0xF8, 0x4D, 0x5B, 0xCF, 0x11,
+				0xA8, 0xFD, 0x00, 0x80, 0x5F, 0x5C, 0x44, 0x2B], {offset: start})) {
+				return {
+					ext: 'wma',
+					mime: 'audio/x-ms-wma'
+				};
+			}
+
+			// ASF_Video_Media  GUID in little endian BC19EFC0-5B4D-11CF-A8FD-00805F5C442B
+			if (check([0xC0, 0xEF, 0x19, 0xBC, 0x4D, 0x5B, 0xCF, 0x11,
```
mifi: Do we need this if we are defaulting to `video/x-ms-wmv` anyhow? (below)

karlhiramoto: My thought was to return from  loop earlier if we've found the WMA `video/x-ms-wmv` GUID. If we removed that line,  functionality would be the same, but would have to complete the loop.

Borewit:
I [wrote an ASF decoder](https://github.com/Borewit/music-metadata/blob/master/src/asf/AsfParser.ts), maybe I should try to extract the bare minimum to determine the type.

PR #139, opened 2018-05-07
Add QCP audio file support

karlhiramoto:
* refactor remove duplicate RIFF header on AVI and WAV
* add QCP https://en.wikipedia.org/wiki/QCP
* add test fixture.qcp


mifi:
Thanks

PR #142, opened 2018-05-30
added support for .ts mpeg transport stream extension

colmaengus:
The .mts support doesn't work for normal .ts files. I'm not sure where the 4 byte offset is coming from.
Rather than change the .mts rule I've created a new one for .ts files.


mifi:
What is this check based on? `check([0x47]) && check([0x47], {offset: 188})`
When i try to match the fixture .ts with unix `file` i just get "data". I thought the thing with TS is that it can start in the middle of a stream so you don't know the offsets, correct me if i'm wrong.

colmaengus:
Hi Mikael,

The sync-byte is certainly used when dealing with streaming content. However for files I think the offset should be zero and then every 188 bytes.

- https://en.wikipedia.org/wiki/List_of_file_signatures
- https://github.com/threatstack/libmagic/blob/master/magic/Magdir/animation#L703

It seems the reason that file is not detecting the fixture.ts as a "MPEG transport stream data" is that the third byte is 0x11 where file is expecting 0x00. For some reason ffmpeg didn't use a PID of 0x00 when generating the .ts file.
I've checked with some other .ts files used in HLS streaming and some use a PID of 0x00 and some don't. So I'm inclined not to enforce this.

Regards

Colm A

mifi:
Will we not risk matching other formats beginning with 0x47, like GIF (0x47=G) which happen to have a 47 at position 188?
I see file's implementation is a bit more strict:
https://github.com/file/file/blob/master/magic/Magdir/animation#L856

sindresorhus:
@colmaengus ⬆️

Issue #149, opened 2018-06-30
WAV file maps to the wrong MIME-type


Borewit:
[WAV](https://nl.wikipedia.org/wiki/WAV) is mapped to `audio/x-wav`, but the recommended MIME-type is `audio/vnd.wave`.

Related: Borewit/music-metadata#110

PR #150, opened 2018-06-30
Fix MIME type for `.wav` and `.avi`

Borewit:
Fix issue #149 

References: 
* https://en.wikipedia.org/wiki/WAV
* https://tools.ietf.org/html/rfc2361

### DIFF
--- a/index.js
+++ b/index.js
@@ -312,13 +312,13 @@ module.exports = input => {
 		if (check([0x41, 0x56, 0x49], {offset: 8})) {
 			return {
 				ext: 'avi',
-				mime: 'video/x-msvideo'
+				mime: 'video/vnd.avi'
 			};
 		}
 		if (check([0x57, 0x41, 0x56, 0x45], {offset: 8})) {
 			return {
 				ext: 'wav',
-				mime: 'audio/wav'
+				mime: 'audio/vnd.wave'
 			};
 		}
 		// QLCM, QCP file
--- a/index.js
+++ b/index.js
@@ -318,7 +318,7 @@ module.exports = input => {
 		if (check([0x57, 0x41, 0x56, 0x45], {offset: 8})) {
 			return {
 				ext: 'wav',
-				mime: 'audio/x-wav'
+				mime: 'audio/wav'
 			};
 		}
 		// QLCM, QCP file


index.js
```diff
@@ -318,7 +318,7 @@ module.exports = input => {
 		if (check([0x57, 0x41, 0x56, 0x45], {offset: 8})) {
 			return {
 				ext: 'wav',
-				mime: 'audio/x-wav'
+				mime: 'audio/wav'
```
sindresorhus: The official MIME-type is: `audio/vnd.wave`

See: https://en.wikipedia.org/wiki/WAV and https://tools.ietf.org/html/rfc2361

sindresorhus: Seems we also need to change `video/x-msvideo` to `video/vnd.avi`

Borewit: You are right!

luckv:
@Borewit @sindresorhus sorry for coming late in the discussion, but I have to say that `audio/vnd.wave` and `video/vnd.avi` are not official mime types, but instead should be used `audio/wav` and `video/avi`, and I tell you why.

Firstly, the RFC you mentioned ([RFC 2361](https://tools.ietf.org/html/rfc2361)), is **not a standard** rfc, but instead an **informational** rfc, that does not represent an Internet community consensus or recommendation.

Secondly, as far as I know, mime types are only used on the web and by browsers to render the content. So, it's logical, to use mime types that browsers know. In my experience with audio files, `audio/vnd.wave` isn't recognized by any major web browser (google chrome, mozilla firefox, safari), but `audio/wav` is.

I have a strong opinion that there isn't an official mime type for .wav and .api files, but in the years emerged an convention to use `audio/wav ` and `video/avi`. If it's not, I will gladly change my opinion.

Here a more long discussion I had on this subject: https://github.com/patrickmccallum/mimetype-io/issues/18

Some sources I have used:
- RFC statuses - https://www.ietf.org/standards/process/informational-vs-experimental/
- RFC 2361 - https://tools.ietf.org/html/rfc2361
- MDN web docs - https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

sindresorhus:
I agree. Practical usage trumps technical specifications in web development. If audio/wav and video/avi work across major browsers and audio/vnd.wave and video/vnd.avi don't, it's clear which MIME types should be used. PR welcome.

luckv:
Ok, I will gladly do the pull request myself.

PR #153, opened 2018-07-09
ASF: WMA or WMV

Borewit:
Address issue #123

**But**: It may not be possible to determine if the file media type (audio or video) within the first 4100 bytes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -330,7 +330,46 @@ module.exports = input => {
 		}
 	}
 
+	function readUInt64LE(buf, offset = 0) {
+		let n = buf[offset];
+		let mul = 1;
+		let i = 0;
+		while (++i < 8) {
+			mul *= 0x100;
+			n += buf[offset + i] * mul;
+		}
+		return n;
+	}
+
+	// ASF_Header_Object first 80 bytes
 	if (check([0x30, 0x26, 0xB2, 0x75, 0x8E, 0x66, 0xCF, 0x11, 0xA6, 0xD9])) {
+		// Search for header should be in first 1KB of file.
+
+		let offset = 30;
+		do {
+			const objectSize = readUInt64LE(buf, offset + 16);
+			if (check([0x91, 0x07, 0xDC, 0xB7, 0xB7, 0xA9, 0xCF, 0x11, 0x8E, 0xE6, 0x00, 0xC0, 0x0C, 0x20, 0x53, 0x65], {offset})) {
+				// Sync on Stream-Properties-Object (B7DC0791-A9B7-11CF-8EE6-00C00C205365)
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -330,10 +341,40 @@ module.exports = input => {
...
+						ext: 'wma',
+						mime: 'application/vnd.ms-asf'
+					};
+				}
+
+				if (check([0xC0, 0xEF, 0x19, 0xBC, 0x4D, 0x5B, 0xCF, 0x11, 0xA8, 0xFD, 0x00, 0x80, 0x5F, 0x5C, 0x44, 0x2B], {offset: offset + 24})) {
+					// Found Audio:
+					return {
+						ext: 'wmv',
+						mime: 'application/vnd.ms-asf'
```
sindresorhus: Shouldn't this be `video/x-ms-wmv`? https://github.com/nginx/nginx/blob/0d224602e966c431674958dabcaa451c99539352/conf/mime.types#L95

index.js
```diff
@@ -330,10 +341,40 @@ module.exports = input => {
...
+		let offset = 30;
+		do {
+			const objectSize = readUInt64LE(buf, offset + 16);
+			if (check([0x91, 0x07, 0xDC, 0xB7, 0xB7, 0xA9, 0xCF, 0x11, 0x8E, 0xE6, 0x00, 0xC0, 0x0C, 0x20, 0x53, 0x65], {offset})) {
+				// Sync on Stream-Properties-Object (B7DC0791-A9B7-11CF-8EE6-00C00C205365)
+				if (check([0x40, 0x9E, 0x69, 0xF8, 0x4D, 0x5B, 0xCF, 0x11, 0xA8, 0xFD, 0x00, 0x80, 0x5F, 0x5C, 0x44, 0x2B], {offset: offset + 24})) {
+					// Found Audio:
+					return {
+						ext: 'wma',
+						mime: 'application/vnd.ms-asf'
```
sindresorhus: This should be `audio/x-ms-wma`. See: https://en.wikipedia.org/wiki/Advanced_Systems_Format

test.js
```diff
@@ -92,9 +92,11 @@ const types = [
 	'xml',
 	'mobi',
 	'heic',
-	'cur',
 	'ktx',
-	'ape'
+	'cur',
+	'ape',
```
sindresorhus: Don't do unrelated changes.

sindresorhus:
You need to add `wma` to the readme.

Borewit:
Pong @sindresorhus, forgot about this PR indeed. I will complete it within a week.

index.js
Borewit: Your source is claiming audio/x-m4a is m4a , which is MPEG-4 layer 3, which should be audio/mp4.

wma and wmv are ASF at the end, I don't understand why folks are confusing things with unregistered MIME-types.

[ASF has an IANA registered MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml#audio), and `audio/x-ms-wma` and `video/x-ms-asf` are not.

I will change it `audio/x-ms-wma` and `video/x-ms-asf`, but it is not that I am very convinced one is better then the other.

PR #156, opened 2018-07-10
issue fix #155

### LINKED ISSUE #155: .xlsx from linux is accepting the file type as  application/zip
i have Uploading the File to the aws i am checking my file type from the buffer 
My buffer Out put value with index
buffer[2] 3
buffer[3] 4
i think it's failing in [see file type code check funciton](https://github.com/sindresorhus/file-type/blob/master/index.js#L193)

>what i need is 
i need to check my xlsx file type using the buffer 
why it's returning .zip format 

-----

[i need to achieve this case ](https://github.com/sindresorhus/file-type/blob/master/index.js#L181)

MuthukumarHelios:
This issue is happening while uploading a xlsx format from linux (ubuntu 16.04)and windows(10)

if we are uploading the xlsx format it is taking it as a zip file format

### DIFF
--- a/index.js
+++ b/index.js
@@ -192,6 +192,19 @@ module.exports = input => {
 	if (
 		check([0x50, 0x4B]) &&
 		(buf[2] === 0x3 || buf[2] === 0x5 || buf[2] === 0x7) &&
+		(buf[3] === 0x4 || buf[3] === 0x6 || buf[3] === 0x8) &&
+		(buf[6] === 0x8)
+	) {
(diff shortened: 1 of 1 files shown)


sindresorhus:
The existing `xlsx` should be fixed instead of just adding a new check.

MuthukumarHelios:
the existing .xlsx is not passing i have tested ...

sindresorhus:
Closing in favor of #159

PR #159, opened 2018-07-25
Fix type detection for `docx`, `xlsx` and `pptx`

### LINKED ISSUE #103: Not correctly recognizing all .docx files
When I extract the file type for the sample .docx file in the unit test it works fine for me.  However when I try extracting the file type for a test file I created on my Mac the library thinks it's a zip file.

[UofTCSCoop.docx](https://github.com/sindresorhus/file-type/files/1289212/UofTCSCoop.docx)

### LINKED ISSUE #119: file-type treat  libreoffice  saved  .pptx  as  .zip
i  just check that,  it is easy to reproduct

### LINKED ISSUE #155: .xlsx from linux is accepting the file type as  application/zip

jacor84:
Fixes #103 
Fixes #155 
Fixes #119 

As I run through multiple docx, xlsx and pptx files, I found that the current implementation was only working sometimes. This is because the order of internal (zip) parts is not fixed, as suggested in the referenced file:
https://github.com/file/file/blob/master/magic/Magdir/msooxml

Example parts order:
```
xl/_rels/workbook.xml.rels
xl/sharedStrings.xml
xl/worksheets/sheet1.xml
xl/workbook.xml
xl/styles.xml
_rels/.rels
docProps/app.xml
docProps/core.xml
[Content_Types].xml
```
So neither `_rels/.rels` nor `[Content_Types].xml` is first, as expected by the current implementation. Instead, I scan the byte array (buffer) until at least one of the mentioned parts and one type-specific part (here: `xl/`) is found.


jacor84:
There is one more thing to consider: because readme says that the lib only needs the first 4100 bytes, this check may not always work for bigger files. It can be easily fixed by sacrificing the necessity to detect one of `_rels/.rels` or `[Content_Types].xml` entries:
```
		while (zipHeaderIndex >= 0) {
			(...)
		}

		// No more zip parts available in the buffer, but maybe we are almost certain about the type?
		if (type) {
			return type;
		}
```

sindresorhus:
> because readme says that the lib only needs the first 4100 bytes, this check may not always work for bigger files

Ugh, I had a feeling that simplicity would be shattered by reality at some point... I guess we could say in the readme that you only need the first 4100 bytes, unless you need to detect the following formats: docx, ...

sindresorhus:
@jacor84 Is your parsing following the recommendation in https://github.com/sindresorhus/file-type/issues/119#issuecomment-356047694 ?

jacor84:
Well, of course it is just a small and quick improvement over existing implementation. I suppose that's the kind of fix that, I guess, bumps .docx recognizability from 70% to >95%. If you include these additional 3 lines of code, then it could even be 100%, maybe risking having a false positive in some well-crafted counter example.

jacor84:
@sindresorhus I pushed the change along with a sample xlsx file in fixtures. It passes the check now, but would not be properly detected by existing algorithm.

sindresorhus:
Thanks, @jacor84 :)

PR #171, opened 2018-11-12
Improve usage example in the readme

m1k1o:
The stream MUST NOT supply enough data to determine what filetype is it. New API allows it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,7 +39,7 @@ const fileType = require('file-type');
 const url = 'http://assets-cdn.github.com/images/spinners/octocat-spinner-32.gif';
 
 http.get(url, res => {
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -39,7 +39,8 @@ const fileType = require('file-type');
 const url = 'http://assets-cdn.github.com/images/spinners/octocat-spinner-32.gif';
 
 http.get(url, res => {
-	res.once('data', chunk => {
+	res.on('readable', () => {
+		let chunk = this.read(fileType.minimumBytes);
```
sindresorhus: You cannot use `this` in an arrow function.

m1k1o: Oh didn't notice. Updated.

PR #172, opened 2018-11-12
Support for all kinds of streams in readme example

m1k1o:
After some feedback and testing I found out, that previous solution works only on sync streams. The `read` function can return null if the buffer is not ready yet. If you want more general solution, this one will work with sync and async streams. Approved by official docs: https://nodejs.org/api/stream.html#stream_readable_read_size


sindresorhus:
Closing for lack of response.

Issue #99, opened 2017-08-15
Many MP4 variants are not implemented


mifi:
https://github.com/file/file/blob/master/magic/Magdir/animation

We should either implement these or finish #68

peterkuiper:
We are handling audio uploads from Android and iOS and `file-type` seems to be returning an incorrect mimetype:

```
file 62513eb2-539e-42ba-af38-56862a1139e1
62513eb2-539e-42ba-af38-56862a1139e1: ISO Media, Apple iTunes ALAC/AAC-LC (.M4A) Audio

file -I 62513eb2-539e-42ba-af38-56862a1139e1
62513eb2-539e-42ba-af38-56862a1139e1: audio/x-m4a; charset=binary

./node_modules/.bin/file-type 62513eb2-539e-42ba-af38-56862a1139e1
m4a
audio/m4a
```
```
file 6ab2b73b-85bb-4384-ba27-8bb92adaaf37
6ab2b73b-85bb-4384-ba27-8bb92adaaf37: ISO Media, MP4 v2 [ISO 14496-14]

file -I 6b93f205-4133-4429-84b8-4b0a9511acc2
6b93f205-4133-4429-84b8-4b0a9511acc2: video/mp4; charset=binary

./node_modules/.bin/file-type 6b93f205-4133-4429-84b8-4b0a9511acc2
mp4
video/mp4
```

We would expect both to be audio/m4a (we are setting the mimetype to `audio/x-m4a`, which seems to work cross platform, but isn't ideal). The `audio/x-m4a` type is what the client (iOS / Android) is detecting.

I'm not sure how to fix this as we are also handling video uploads. The `video/mp4` type is quite confusing when we are processing an audio file.

IssuehuntBot:
@issuehunt has funded $80.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/19105307/issues/99) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/19105307/issues/99).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

IssuehuntBot:
@sindresorhus has rewarded $72.00 to @stroncium. [See it on IssueHunt](https://issuehunt.io/repos/19105307/issues/99)

[...]

Issue #60, opened 2016-08-03
Not all mp4 are recognized


sam1111:
Hi,

I am attaching an mp4 file which is not getting recognized correctly. the mimetype returned by module comes as undefined.
[videoplayback.mp4.zip](https://github.com/sindresorhus/file-type/files/398775/videoplayback.mp4.zip)

t1st3:
Getting some info about the file with [mediainfo](https://mediaarea.net/en/MediaInfo), I get this kid of output:
```
General
Complete name                            : videoplayback.mp4
Format                                   : dash
Codec ID                                 : dash
File size                                : 4.57 MiB
```

I don't know much about `mp4` files, but it appears that `dash` maybe something else but related to `mp4`. Wikipedia has 2 separate pages about `mp4` and `dash`:
* [mp4](https://en.wikipedia.org/wiki/MPEG-4_Part_14)
* [dash](https://en.wikipedia.org/wiki/Dynamic_Adaptive_Streaming_over_HTTP)

Seems that `dash` is a different codec from traditional `mp4` while keeping the same extension.

mifi:
I have added DASH as mp4 (as `file` identifies it).

65f0017d8a418dd217e8eec9250c2d9f04578f56

However there are still many other mp4 variants not yet detected by file-type, see #99

Issue #38, opened 2015-11-04
not all mp4s are recognized


davibe:
Hi,
I got an mp4 file that is not recognized. This is the first 262 bytes. I hope this is useful.
https://dl.dropboxusercontent.com/u/968175/test.mp4

sindresorhus:
Are you sure it's not a `.mov` or something with a renamed extension? I've seen that before. Just the first first bytes isn't very useful for being able to debug what the issue is, though. Would you be able to submit a second or less video clip of anything with the same issue? Preferably as a pull request with a failing test.

davibe:
I have many of them but i can't share.. Does this help ?

```
Last login: Thu Nov  5 10:23:38 on ttys003
dade@choo:~$ ffmpeg -i /Users/dade/work/asd.mp4 
ffmpeg version 2.8.1 Copyright (c) 2000-2015 the FFmpeg developers
  built with Apple LLVM version 7.0.0 (clang-700.1.76)
  configuration: --prefix=/usr/local/Cellar/ffmpeg/2.8.1_1 --enable-shared --enable-pthreads --enable-gpl --enable-version3 --enable-hardcoded-tables --enable-avresample --cc=clang --host-cflags= --host-ldflags= --enable-opencl --enable-libx264 --enable-libmp3lame --enable-libvo-aacenc --enable-libxvid --enable-librtmp --enable-libfdk-aac --enable-nonfree --enable-vda
  libavutil      54. 31.100 / 54. 31.100
  libavcodec     56. 60.100 / 56. 60.100
  libavformat    56. 40.101 / 56. 40.101
  libavdevice    56.  4.100 / 56.  4.100
  libavfilter     5. 40.101 /  5. 40.101
  libavresample   2.  1.  0 /  2.  1.  0
  libswscale      3.  1.101 /  3.  1.101
  libswresample   1.  2.101 /  1.  2.101
  libpostproc    53.  3.100 / 53.  3.100
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '/Users/dade/asd.mp4':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomiso2mp41
    creation_time   : 1970-01-01 00:00:00
    encoder         : Lavf52.93.0
  Duration: 00:04:16.54, start: 0.000000, bitrate: 1151 kb/s
[...]

PR #177, opened 2018-11-28
Add support for MPEG-4 Audiobook (.m4b)

### LINKED ISSUE #165: m4b audio books maps to video/mp4
Ref: 
- https://en.wikipedia.org/wiki/MPEG-4_Part_14
- https://en.wikipedia.org/wiki/Audio_file_format

Related issues:
- #147 _MPEG-4 Audio maps to wrong MIME-type_
- #122: _m4a from android phone_
- #99: _Many MP4 variants are not implemented_

Borewit:
Fixes #165
Related: #99, #60, #38

Note that I removed `check([0x4D, 0x34, 0x41, 0x20])`, at offset 0. I don't think this picking out any MPEG-4 audio file.


sindresorhus:
Tests are failing as the audiobook fixture is being detected as `.mp4`.

Borewit:
Not sure if it is actually that simple to pick out the `.m4b` extension.

sindresorhus:
@Borewit Still interested in finishing this? :)

Borewit:
You may close the issue.
I don't think there is a feasible way to extract the audio-book type.
Although this can be found in the metadata, the complexity to get that out with the risk it might not be within the first 4k, I don't think it's worth to implement it.

I started this PR based on the assumption it could be extracted with the first bytes, which appears not to be the case.

I went through the changes of this PR, and kept a minor code correction, which I moved to [PR](https://github.com/sindresorhus/file-type/pull/184).

I use this module for [music-metadata](https://github.com/Borewit/music-metadata) to determine the audio file type, based on the content, as a fallback in case the audio type wasn't provided by the user. As long as I get one of the  MPEG-4 MIME-types it works for me.

Feel free to ping me on any discussion regarding audio file type, or on requests to cover new audio file types.

sindresorhus:
I understand. No worries at all :)

> Feel free to ping me on any discussion regarding audio file type, or on requests to cover new audio file types.

Will do.

PR #187, opened 2019-02-22
Add `.stream()` method

### LINKED ISSUE #186: Add method that returns Node stream back to normal state
I can PR this later - it'd be nice to have a method that returns the stream back to a usable state in case we want to keep using the stream.

Currently, this is what I do:

```js
function getFileStreamWithType() {
    const chunk = readStream.read(fileType.minimumBytes);
    const thisFileType = fileType(chunk);

    readStream.removeListener('readable', getFileStreamWithType);
    readStream.unshift(chunk);

    callback(null, thisFileType, readStream);
}

readStream.on('readable', getFileStreamWithType);

function callback(err, thisFileType, readStream) {
    readStream.pipe(request.put('example.com'));
    ...
});
```

Maybe something like:

```js
const [thisFileType, usableStream] = fileType.getStream(readableStream);
```

bencmbrook:
This PR adds a method to support Node streams without causing side effects on the state of the stream. It reads in the minimum amount of bytes to detect the filetype, detects the file type, then rolls back the stream for use elsewhere.

Fixes #186

### DIFF
--- a/index.js
+++ b/index.js
@@ -919,16 +919,15 @@ module.exports = input => {
 
 Object.defineProperty(module.exports, 'minimumBytes', {value: 4100});
 
-module.exports.stream = readableStream => new Promise((resolve, reject) => {
+module.exports.stream = readableStream => new Promise(resolve => {
 	const stream = eval('require')('stream'); // eslint-disable-line no-eval
 
 	readableStream.once('readable', () => {
 		const pass = new stream.PassThrough();
-		const chunk = readableStream.read(module.exports.minimumBytes);
+		const chunk = readableStream.read(module.exports.minimumBytes) || readableStream.read();
 		pass.fileType = module.exports(chunk);
 		readableStream.unshift(chunk);
 
-		// TODO: Use `stream.pipeline()` when targeting Node.js 10
 		resolve(readableStream.pipe(pass));
 	});
 });
--- a/index.js
+++ b/index.js
@@ -928,6 +928,7 @@ module.exports.stream = readableStream => new Promise(resolve => {
 		pass.fileType = module.exports(chunk);
 		readableStream.unshift(chunk);
 
+		// TODO: Use `stream.pipeline()` when targeting Node.js 10
 		resolve(readableStream.pipe(pass));
 	});
 });
--- a/index.js
+++ b/index.js
@@ -920,19 +920,15 @@ module.exports = input => {
 Object.defineProperty(module.exports, 'minimumBytes', {value: 4100});
 
 module.exports.stream = readableStream => new Promise((resolve, reject) => {
-	try {
-		const stream = eval('require')('stream'); // eslint-disable-line no-eval
+	const stream = eval('require')('stream'); // eslint-disable-line no-eval
 
-		readableStream.once('readable', () => {
-			const pass = new stream.PassThrough();
-			const chunk = readableStream.read(module.exports.minimumBytes);
-			pass.fileType = module.exports(chunk);
-			readableStream.unshift(chunk);
-			
-			// TODO: Use `stream.pipeline()` when targeting Node.js 10
-			resolve(readableStream.pipe(pass));
-		});
-	} catch (error) {
-		reject(error);
-	}
+	readableStream.once('readable', () => {
+		const pass = new stream.PassThrough();
+		const chunk = readableStream.read(module.exports.minimumBytes);
+		pass.fileType = module.exports(chunk);
+		readableStream.unshift(chunk);
+
+		// TODO: Use `stream.pipeline()` when targeting Node.js 10
+		resolve(readableStream.pipe(pass));
+	});
 });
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -918,3 +918,16 @@ module.exports = input => {
...
+
+	function helper(stream) {
+		const chunk = stream.read(module.exports.minimumBytes);
+		stream.fileType = module.exports(chunk);
+		stream.unshift(chunk);
+		resolve(stream);
+	}
+
+	stream.once('readable', () => helper(stream));
+});
```
sindresorhus: This could be written simpler without the `helper` function:

```js
module.exports.stream = readableStream => new Promise(resolve => {
	readableStream.once('readable', () => {
		const chunk = readableStream.read(module.exports.minimumBytes);
		readableStream.fileType = module.exports(chunk);
		readableStream.unshift(chunk);
		resolve(readableStream);
	});
});
```

test.js
```diff
@@ -3,6 +3,8 @@ import test from 'ava';
 import readChunk from 'read-chunk';
 import fileType from '.';
 
+const fs = require('fs');
```
sindresorhus: Use `import`

sindresorhus:
Travis is failing

index.js
bencmbrook: Agreed. It was initially there because of the `.removeListener` method. With `.once` it's not necessary.

bencmbrook:
> Travis is failing

Interesting. Will look into it.

bencmbrook:
Piping the `readableStream` through a PassThrough stream is a cleaner way of doing this because it preserves the event lifecycle and the stream flowing mode. Before, it was switching modes, which was causing it to break it Node 10.

This does require that we import the `stream` node module in `index.js`. I didn't use the ES6 module syntax there.

The open question then becomes, how do we want to keep this browser compatible? The stream method nor the require should be used in browsers.

sindresorhus:
> Piping the `readableStream` through a PassThrough stream is a cleaner way of doing this because it preserves the event lifecycle and the stream flowing mode. Before, it was switching modes, which was causing it to break it Node 10.

Agreed

> The open question then becomes, how do we want to keep this browser compatible? The stream method nor the require should be used in browsers.

Webpack is super annoying in this sense and tries to import every require, even if it's behind an if-statement or not even used. One way to work around this is to use `eval`:

```js
module.exports.stream = () => {
	const stream = eval('require')('stream');
	// …
};

File: index.js [hunk 3]
```diff
@@ -924,6 +921,8 @@ Object.defineProperty(module.exports, 'minimumBytes', {value: 4100});
 
 module.exports.stream = readableStream => new Promise((resolve, reject) => {
 	try {
```
sindresorhus: You don't need to try/catch here. Any errors thrown in the `Promise` constructor makes the promise reject.

readme.md
```diff
@@ -87,6 +87,16 @@ Type: `number`
 
 The minimum amount of bytes needed to detect a file type. Currently, it's 4100 bytes, but it can change, so don't hardcode it.
 
+### fileType.stream(readableStream)
+
+Detect the file type of a readable stream.
+
+Returns a `Promise` which resolves to the original readable stream argument, but with an added `fileType` property, which is an object like the one returned from `fileType()`.
+
```
sindresorhus: Can you include a quick example? For example: https://github.com/sindresorhus/file-type/pull/187/files#diff-1dd241c4cd3fd1dd89c570cee98b79ddR230

test.js
```diff
@@ -223,3 +224,30 @@ test('validate the input argument type', t => {
...
+	const readableStream = await fileType.stream(fs.createReadStream(file));
+
+	t.deepEqual(readableStream.fileType, fileType(readChunk.sync(file, 0, fileType.minimumBytes)));
+});
+
+test('check identical stream', async t => {
+	const file = path.join(__dirname, 'fixture', 'fixture.mp3');
+
+	const readableStream = await fileType.stream(fs.createReadStream(file));
+	const bufA = [];
```
sindresorhus: `bufA` => `bufferA`

test.js
```diff
@@ -223,3 +224,30 @@ test('validate the input argument type', t => {
...
+
+	const fileStream = fs.createReadStream(file);
+	const bufB = [];
+
+	readableStream.on('data', c => bufA.push(Buffer.from(c)));
```
sindresorhus: `c` => `chunk`

readme.md
bencmbrook: Yup for sure

bencmbrook:
Bug found on the `.rtf` file. `.read` returns null. Should test all files.

https://github.com/bencmbrook/file-type/blob/929de6c441bc9b0c2cb6642f15a3d239ee7d6470/index.js#L928

bencmbrook:
Features are done and tests are all passing. Anything you want changed in the latest commits?

bencmbrook:
Marked everything as resolved

bencmbrook:
🤙sweet. I'm gonna get a lot of use out of this.

Also, minor docs error on that readme commit: https://github.com/sindresorhus/file-type/commit/e8d4e1f2551d83224a9b3be2d671979e165aaceb#r32592498

intellix:
PassThrough causes issues with rewind for me: https://github.com/sindresorhus/file-type/issues/216

bencmbrook:
How so? What version of file-type are you using?

PR #188, opened 2019-03-05
Add TypeScript definition

BendingBender:


index.d.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @returns An object with the detected file type and MIME type or `null` when there was no match.
+ */
+export default function fileType(
+	buffer: Buffer | Uint8Array
+): FileTypeResult | null;
+
+/**
+ * The minimum amount of bytes needed to detect a file type. Currently, it's 4100 bytes, but it can change, so don't hard-code it.
+ */
+export const minimumBytes: number;
```
sindresorhus: This one should be a property on `fileType` and no a named export.

index.d.ts
```diff
@@ -0,0 +1,139 @@
+/// <reference types="node"/>
+import {Readable} from 'stream';
```
sindresorhus:
```suggestion
import {Readable as ReadableStream} from 'stream';
```

To make it more readable. It's not clear further down in the code where it's used what `Readable` is.

PR #195, opened 2019-04-02
Handle errors from `fileType.stream` method

chytreg:
There are cases like an empty stream then fileType(chunk) will throw an error. The error is not propagated correctly and cannot be caught in the main plot. Using reject solves the problem and allows to handle error in the main plot of the app.
```js
try {
  fileTypeStream = await fileType.stream(stream);
} catch (error) {
  // without reject, you won't catch the error thrown by fileType(check) here
  console.log(error);
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -941,6 +941,7 @@ module.exports.stream = readableStream => new Promise((resolve, reject) => {
 		} catch (error) {
 			reject(error);
 		}
+
 		readableStream.unshift(chunk);
 
 		if (stream.pipeline) {
--- a/index.js
+++ b/index.js
@@ -929,14 +929,18 @@ module.exports.default = fileType;
 
 Object.defineProperty(fileType, 'minimumBytes', {value: 4100});
 
-module.exports.stream = readableStream => new Promise(resolve => {
+module.exports.stream = readableStream => new Promise((resolve, reject) => {
 	// Using `eval` to work around issues when bundling with Webpack
 	const stream = eval('require')('stream'); // eslint-disable-line no-eval
 
 	readableStream.once('readable', () => {
 		const pass = new stream.PassThrough();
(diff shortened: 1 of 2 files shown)


sindresorhus:
@bencmbrook Thoughts?

bencmbrook:
@sindresorhus this is a solid PR. I tested it independently and this does improve the stack trace. It doesn't change the API per se, as it will still throw an error in a similar manner, but if it's uncaught at the top-level, this PR will make it easier to debug.

@chytreg might consider adding a more verbose error describing the empty stream.

chytreg:
@bencmbrook I think we may try to improve a message for TypeError of fileType method.

```js
const fileType = input => {
	if (!(input instanceof Uint8Array || Buffer.isBuffer(input))) {
		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\`, got \`${typeof input}\``);
	}
```

typeof(null) => "object",
typeof(undefined) => "undefined"

```js
const fileType = input => {
	if (!(input instanceof Uint8Array || Buffer.isBuffer(input))) {
                const inputTypeMsg = (input) ? typeof(input) : "an empty input";
		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\`, got \`${inputTypeMsg}\``);
	}
```

chytreg:
We can handle null also as follow:
```js
const fileType = input => {
        if (!input) { return null }
	if (!(input instanceof Uint8Array || Buffer.isBuffer(input))) {
                const inputTypeMsg = (input) ? typeof(input) : "an empty input";
		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\`, got \`${inputTypeMsg}\``);
	}

```

bencmbrook:
@chytreg isn't this essentially what would get thrown when it reads a chunk anyway?

PR #198, opened 2019-04-09
Fix Node.js 11 compatibility

jfinkhaeuser:
Fixes #196

PR #199, opened 2019-04-09
Run travis CI also on Node.js 11

fstoerkle:
As of https://github.com/sindresorhus/file-type/pull/198, file-type should also work on Node.js 11.
Running the tests on Node.js 11 would verify that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - '11'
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks, but I don't test on unstable Node.js versions. I did that in the past and it caused too many problems. Those versions often break. It has been verified locally. That's enough.

PR #200, opened 2019-04-11
Add support for `lnk` and `alias` formats

### LINKED ISSUE #192: Question: how to detect if a file is an "alias" or a "shortcut" of a file ?

stroncium:
https://forensicswiki.org/wiki/LNK
https://en.wikipedia.org/wiki/Alias_(Mac_OS)#File_structure

Fixes #192


stroncium:
solves #192

index.js
```diff
@@ -920,6 +920,20 @@ const fileType = input => {
 		};
 	}
 
+	if (check([0x4C, 0x00, 0x00, 0x00, 0x01, 0x14, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46])) {
+		return {
+			ext: 'lnk',
+			mime: 'application/x.ms.shortcut'
```
sindresorhus: This is not the correct MIME type.

index.js
```diff
@@ -920,6 +920,20 @@ const fileType = input => {
...
+		};
+	}
+
+	if (check([0x62, 0x6F, 0x6F, 0x6B, 0x00, 0x00, 0x00, 0x00, 0x6D, 0x61, 0x72, 0x6B, 0x00, 0x00, 0x00, 0x00])) {
+		return {
+			ext: 'alias',
+			mime: 'application/x.apple.alias'
```
sindresorhus: I tried Gooling this without any results.

sindresorhus:
Use the `Fixes #192` wording in the PR description (applies to all your PRs), so the issue is automatically closed when the PR is merged.

index.js
stroncium: There is no mention of specific mime type for them anywhere I tried.
x mime types are not registered and don't signal anything by their format.
I used scheme people usually apply to Microsoft and Apple related formats.

`file` gives `application/octet-stream`for them, but for the purposes of say Content-Type it isn't correct, it should be missing instead.

So, there are 3 options:
- using x tree and coming up with new type, at least for the people who just check it programmatically. It will also be correct for supplying to other applications, as it isn't registered nor known type.
- using `undefined`. It is correct for the purposes of Content-Type, but doesn't help with programmatic use, doesn't happen anywhere in the library and changes type definitions for typescript.
- using `application/octet-stream`, like `file`, which is incorrect and might trigger unwanted behavior in browsers for example.
- using `application/unknown` which is correct, but doesn't signal anything to api users, just as `undefined`

index.js
stroncium: Above.

index.js
sindresorhus: The one mention in the link is using dashes. You're using `.`

sindresorhus:
```suggestion
			mime: 'application/x-ms-shortcut'
```

stroncium: Using `x-` is against modern standards.
'.' is recommended delimiter for parts of subtype.
Unregistered tree is specifically defined by `x.`, not `x-` nor just `x`.

sindresorhus: Ah ok. I didn't know that.

sindresorhus: Can you add a comment that the MIME is just invented by us? Same with the other one.

stroncium: Well, technically, `x.` already indicates it was just invented by random person. I'll add it to readme.

sindresorhus: Yes, but sometimes it's a common convention shared by multiple tools/software. Here, it's just *us* using it.

PR #202, opened 2019-04-12
Check that MIME type is present in the tests

### LINKED ISSUE #133: Tests do not cover mime-type
Hi,

Very useful package, however looking through the tests for it I notice that you only test for the correct extension being returned for each file and do not check mime type. Jut wondered is there a reason for that?

alexzherdev:
Fixes #133

### DIFF
--- a/test.js
+++ b/test.js
@@ -8,8 +8,7 @@ import fileType from '.';
 
 const check = (ext, name) => {
 	const file = path.join(__dirname, 'fixture', `${(name || 'fixture')}.${ext}`);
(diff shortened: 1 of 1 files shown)


test.js
```diff
@@ -195,7 +194,9 @@ const names = {
 };
 
 const testFile = (t, type, name) => {
-	t.is(check(type, name), type);
+	const {ext, mime} = check(type, name);
+	t.is(ext, type);
+	t.truthy(mime);
```
sindresorhus: To be entirely correct, this should test that it's a string.
```suggestion
	t.is(typeof mime, 'string');
```

PR #204, opened 2019-04-14
Add support for pre-POSIX `tar` 

### LINKED ISSUE #76: Pre-POSIX.1-1988 (i.e. v7) tar header
Hi, is it possible to add the headers for Pre-POSIX.1-1988 (i.e. v7) tar recognition?

Reference here: https://en.wikipedia.org/wiki/Tar_(computing)#cite_note-2

Thanks in advice!

stroncium:
Fixes #76 

Replaced ustar magic check with checksum magic check and checksum validation.
Attempts to add some more magic failed, tar implementations don't seem to respect `pad wtih NUL` policy even for legacy mode.
Recommendations from https://en.wikipedia.org/wiki/Tar_(computing)


index.js
```diff
@@ -16,6 +16,31 @@ function readUInt64LE(buf, offset = 0) {
...
+		return false;
+	}
+
+	let sum = 256;
+	let signedBitCount = 0;
+
+	for (let i = 0; i < 148; i++) {
+		const byte = buf[i];
+		sum += byte;
+		signedBitCount += byte & 0x80;
```
sindresorhus: Can you move `0x80` to a variable with a descriptive name?

index.js
```diff
@@ -16,6 +16,31 @@ function readUInt64LE(buf, offset = 0) {
...
+	for (let i = 156; i < 512; i++) {
+		const byte = buf[i];
+		sum += byte;
+		signedBitCount += byte & 0x80;
+	}
+
+	const readSum = parseInt(buf.toString('utf8', 148, 154), 8);
+
+	return sum === readSum || ((sum - (signedBitCount << 1)) === readSum);
+};
```
sindresorhus: Can you add some comments in this function to explain what's going on?

index.js
stroncium: But it's not a variable, it's bitmask.

stroncium: done

index.js
stroncium: done

sindresorhus:
Thanks. Looks good now :)

index.js
```diff
@@ -16,6 +16,37 @@ function readUInt64LE(buf, offset = 0) {
...
+		sum += byte; // Add to sum
+		signedBitSum += byte & MASK_8TH_BIT; // Add signed bit to signed bit sum
+	}
+
+	// Skip checksum field
+
+	for (let i = 156; i < 512; i++) {
+		const byte = buf[i];
+		sum += byte; // Add to sum
+		signedBitSum += byte & MASK_8TH_BIT; // Add signed bit to signed bit sum
```
noisyui: Clear all lower bits except for the signed bit?

PR #223, opened 2019-06-30
Doc xls ppt support

leizhao175:
Hi,

I have added an extension to your lib to support .**doc**, .**xls**, and .**ppt** (97 ~ 2003) version.

The logic is based on http://fileformats.archiveteam.org/wiki/Microsoft_Compound_File.

As you may know, to identify these file types basically requires loading the whole file into chunk, therefore, I have to make a little change on the interface. But I tried my best to minimise the impact close to zero.

Please let me know if you like it, or if I missed anything.

Here are my modifications:

**fileType(chunk)**
- If the chunk is not a 'msi' file, no change to the response.
- If the chunk is one of **doc**, **xls**, **ppt**, and it is **LARGE** enough to identify the type, it will return:
`{ext: 'doc',mime: 'application/msword'}`
`{ext: 'xls',mime: 'application/vnd.ms-excel'}`
`{ext: 'ppt',mime: 'application/vnd.ms-powerpoint'}`
- If the chunk is one of **doc**, **xls**, **ppt**, and it is not large enough to identify the type, it will return:
`{ext: 'msi', mime: 'application/x-msi', minimumRequiredBytes: 1234567}`
Which is almost same as existing response, with an extra field '_minimumRequiredBytes_', it tells user how large the chunk is required to identify this file. And if user wants, he / she can re-invoke the function with a larger chunk.

**fileType.stream()**
- I have modified this function so it can automatically load extra chunk from the stream if it's required.
- This stream still returns a PassThrough stream with extra field _fileType_ as it is
- If the stream is a .doc, .xls, .ppt, this fun will return the right type as listed above to _fileType_
- No impact to users, as this function still accept and return a stream, no change.

**testFile() & testFileFromStream()**
- I have to light modify these two functions in _test.js_ so they can identify the new field '_minimumRequiredBytes_', and provide a larger chunk for testing.

**Impact to users:**
- For users using the **stream** function, there is no impact to them.
- For users using the **chunk** function, and if they don't care '.doc | .xls | .ppt', there is no impact to them as well, as they will get the same result as it is with an extra field '_minimumRequiredBytes_', which they should just ignore it.
- For users using the **chunk** function, and they do care '.doc | .xls | .ppt'. They must be new users as it was not available. And we can help them now :)


sindresorhus:
While I appreciate the PR, I'm not really interested in supporting these ancient formats. Especially not when they cause lots of churn to the code base. I would recommend making a separate module that only checks for those formats and could be used in combination with `file-type`, or a fork of `file-type`.

Borewit:
@ajayvignesh01, I created this repo: https://github.com/Borewit/file-type-cfbf

You mind to add your code there as a PR: 

If you can create a commit like this: https://github.com/Borewit/file-type-cfbf/commit/e8fb7af16c8c9ad4881b6846eeefcabac843f069

That way I can make sure you are properly credited as the original author.

ajayvignesh01:
> @ajayvignesh01, I created this repo: https://github.com/Borewit/file-type-cfbf
> 
> 
> 
> You mind to add your code there as a PR: 
> 
> 
> 
> If you can create a commit like this: https://github.com/Borewit/file-type-cfbf/commit/75fbb7d3df9f86e00ceaeeda6f71f6b1e9b41ea0
> 
> 
> 
> That way I can make sure you are properly credited as the original author.
> 
> 

Sure! On vacation, can get to it next week

Borewit:
> Sure! On vacation, can get to it next week

A kind reminder @ajayvignesh01

PR #227, opened 2019-07-10
Improve performance of the ZIP check

### LINKED ISSUE #226: 100% cpu with 177.3 MiB zip
I accidentally passed a bigger buffer to fileType, 
i.e. this file https://imagemagick.org/download/windows/releases/ImageMagick-6.9.10-33.zip

Seems to hang iterating through the whole buffer here: 
https://github.com/sindresorhus/file-type/blob/a1da90b2c1121377e47f5dec1f1ce263b5b4f31f/index.js#L275
Is that expected?

It doesn't hang of corse if I pass just a subarray 
`filetype(body.subarray(0, filetype.minimumBytes))` then everything works and `{ ext: 'zip', mime: 'application/zip' }` is returned

jacor84:
Closes #226

Changed `findNextZipHeaderIndex` function to use one `Buffer.indexOf` call instead of `Uint8Array.findIndex`, which calls the passed function for each byte. For this to work, I changed the `buffer` variable type to `Buffer`. It extends `Uint8Array`, but it does not copy data, so it does not have any memory impact.

Btw, there is `indexOf` method on `Uint8Array` type, but it only allows for single byte searching, not a byte sequence.

Tested on this file https://imagemagick.org/download/windows/releases/ImageMagick-6.9.10-33.zip
Result: processing time drop from several hours (I stopped it after 90 minutes) to 84 milliseconds.
My setup: MacBook Pro, Intel Core I7 2,2 GHz, node 12.5.0.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ const fileType = input => {
 		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\` or \`ArrayBuffer\`, got \`${typeof input}\``);
 	}
 
-	const buffer = input instanceof Uint8Array ? input : new Uint8Array(input);
+	const buffer = Buffer.from(input);
 
 	if (!(buffer && buffer.length > 1)) {
 		return;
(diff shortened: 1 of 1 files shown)


sindresorhus:
Wow. Very nice!

cafeoh:
Correct me if I'm wrong, but doesn't this change breaks this module for browsers?

We've been using it until before this PR just fine, since Uint8Array isn't node-specific, but since `Buffer` is now required, we have to install the `buffer` package (which isn't listed as a dependency) and also add `Buffer` to the globals (since it's not imported anywhere).

jacor84:
@cafeoh Please take a look at [this PR](https://github.com/sindresorhus/file-type/pull/232) and check if it solves your problem.

PR #232, opened 2019-08-03
Make Buffer usage optional again

jacor84:
Related to #227 and [a comment about browser usage](https://github.com/sindresorhus/file-type/pull/227#issuecomment-517741187).

This PR makes usage of `Buffer` optional. If input type is `Buffer`, zip header searching is conducted with much faster `Buffer.indexOf`, otherwise single-byte `Uint8Array.indexOf` is used, which results in many false positives that must be discarded. Thus using `Buffer` type as input is still the best option in terms of speed and memory usage.

Tested on [the same 169MB .zip file as before](https://imagemagick.org/download/windows/releases/ImageMagick-6.9.10-33.zip), resulting in 49ms with `Buffer` and 465ms without `Buffer`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,9 @@
 'use strict';
-const {stringToBytes, readUInt64LE, tarHeaderChecksumMatches, uint8ArrayUtf8ByteString} = require('./util');
+const {multiByteIndexOf, stringToBytes, readUInt64LE, tarHeaderChecksumMatches, uint8ArrayUtf8ByteString} = require('./util');
 
 const xpiZipFilename = stringToBytes('META-INF/mozilla.rsa');
 const oxmlContentTypes = stringToBytes('[Content_Types].xml');
 const oxmlRels = stringToBytes('_rels/.rels');
-const zipHeader = [0x50, 0x4B, 0x3, 0x4];
 
 const fileType = input => {
 	if (!(input instanceof Uint8Array || input instanceof ArrayBuffer || Buffer.isBuffer(input))) {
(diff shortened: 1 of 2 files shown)


jacor84:
@sindresorhus What do you think about it?

index.js
```diff
@@ -237,7 +237,23 @@ const fileType = input => {
...
+			while (index >= 0) {
+				if (check(zipHeader, {offset: index})) {
+					return index;
+				}
+
+				index = buffer.indexOf(zipHeader[0], index + 1);
+			}
+
+			return -1;
+		};
```
sindresorhus: Can you make this a generic utility function and move it to https://github.com/sindresorhus/file-type/blob/master/util.js ? Could be called `multiByteIndexOf` or something better.

index.js
```diff
@@ -4,13 +4,14 @@ const {stringToBytes, readUInt64LE, tarHeaderChecksumMatches, uint8ArrayUtf8Byte
 const xpiZipFilename = stringToBytes('META-INF/mozilla.rsa');
 const oxmlContentTypes = stringToBytes('[Content_Types].xml');
 const oxmlRels = stringToBytes('_rels/.rels');
+const zipHeader = [0x50, 0x4B, 0x3, 0x4];
```
sindresorhus: Can you revert this change? Doesn't really make sense here as it's not used anywhere else.

index.js
jacor84: 👍

PR #236, opened 2019-08-26
add plugin support

echosoar:
## Description

Complex type judgments may be used by very few people, and add them into the main code can lead to performance degradation for most users.

This PR adds a method that make user to check complex type by plugin.

## Demo

```
fileType.plugin('test', (_, ctx) => {
	if (ctx.check([0x87, 0x88, 0x89, 0x90])) {
		return {
			ext: 'test',
			mime: 'application/test'
		};
	}
});
const result = fileType(new Uint8Array([0x87, 0x88, 0x89, 0x90]));
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const supported = require('./supported');
 const xpiZipFilename = stringToBytes('META-INF/mozilla.rsa');
 const oxmlContentTypes = stringToBytes('[Content_Types].xml');
 const oxmlRels = stringToBytes('_rels/.rels');
+const plugins = new Map();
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Complex type judgments may be used by very few people, and add them into the main code can lead to performance degradation for most users.

Can you provide an example of such type?

---

I think it would be more useful with a generic hook instead, which could add one or more types, or block a type. Don't forget that adding extra types needs to update `fileType.extensions` and `fileType.mimeTypes` too. Same with the TypeScript types.

Issue #233, opened 2019-08-15
Add tests to ensure all the data is in sync


sindresorhus:
See https://github.com/sindresorhus/file-type/issues/91#issuecomment-519347340.

PR #237, opened 2019-08-27
Add test to validate that all extensions/mimes are in sync across files

bencmbrook:
Closes https://github.com/sindresorhus/file-type/issues/233

Tests for:
1. Duplicates: extensions/mimes that are defined twice in a file
2. Extras: extensions/mimes that are in another file but not in `index.js`
3. Missing: extensions/mimes that are in `index.js` but missing from another file

This treats `index.js` as the base source of truth for which extensions are implemented (as it should).

The tests found a bunch of duplicates and missing extensions, so there are some fixes in those files as well to make the tests actually pass :)

Note: `.asf`, `.ogm`, `.ogx`, and `.mpc`  don't have test files. Since this is out of scope of my PR, I've declared them as exceptions with the `missingTests` field.


sindresorhus:
This is really great. Thanks for the high-quality PR 👍

sindresorhus:
@bencmbrook Would you be interested in being a maintainer on this project? You've done good work here. No worries if not. Just thought I'd ask :)

bencmbrook:
@sindresorhus Thanks! I'm a big fan of the project, and would be pleased to become a maintainer.

sindresorhus:
Awesome :)

PR #238, opened 2019-08-27
docs(README.md): extensions sorted and lowercased

martingronlund:
Only uppercase extension was `Z` - now `z`


sindresorhus:
I'm willing to do it if you add a test that enforces it. Similar to https://github.com/sindresorhus/file-type/blob/63567c8d102607d9119d2ebe3c8465a1673ebb97/test.js#L266.

Issue #241, opened 2019-09-05
Check archive types before any other type


AshtonRoute:
Thanks for the great library.

Unfortunately, I stumbled upon a very ~funny~ serious bug in our environment.

We work with archives a lot, mainly TAR archives.

If you look at the TAR archive structure you can see that it starts with an inode name.
And if you take a look at `index.js`, for instance. You can see that you check some file types before you check that it's actually an archive.

```javascript
if (check([0x42, 0x4D])) {
  return {
    ext: 'bmp',
    mime: 'image/bmp'
  };
}
```

Which means, that it might report a wrong file type if it matches any of the file type before TAR type check.

For instance, one might create a folder named *BMW*, pack it into a TAR archive and the library will report that it's a BMP image.

I'm working on a PR right now

P.S.

I wonder how come no one noticed this before :disappointed: 

[Update]

PR #242

PR #242, opened 2019-09-05
Check tar type before any other type

AshtonRoute:
Fixes sindresorhus/file-type#241


.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 yarn.lock
+.idea
```
kevva: These are unrelated. Please remove.

AshtonRoute:
@kevva 

>Could you add a test for this as well?

I did.

https://github.com/sindresorhus/file-type/raw/2a9d3c26d7da81623809718b17e58a0150e5f5cf/fixture/fixture.tar

kevva:
> I did.

Sorry, my bad. Did not spot you changing the fixture file :).

.gitignore
AshtonRoute: Done

AshtonRoute:
Another consideration:
```javascript
if (
    check([0x30, 0x30, 0x30, 0x30, 0x30, 0x30], {offset: 148, mask: [0xF8, 0xF8, 0xF8, 0xF8, 0xF8, 0xF8]}) && // Valid tar checksum
    tarHeaderChecksumMatches(buffer)
) {
    return {
        ext: 'tar',
        mime: 'application/x-tar'
    };
}
```

Right now this function returns 'tar' only if both Magic Number and the checksum are valid.

I suggest that we check for Magic Number first
and then if the checksum is valid return 'tar'
Otherwise throw an error (breaking change) or return null

index.js
```diff
@@ -46,6 +46,7 @@ const fileType = input => {
 
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
+	// we have to check for tar type before any other type
```
kevva: I'd either clarify why or remove this comment altogether.

sindresorhus:
@AshtonRoute Agreed. Make it:

```js
if (
	check([0x30, 0x30, 0x30, 0x30, 0x30, 0x30], {offset: 148, mask: [0xF8, 0xF8, 0xF8, 0xF8, 0xF8, 0xF8]})
) {
	if (!tarHeaderChecksumMatches(buffer)) {
		return;
	}

	return {
		ext: 'tar',
		mime: 'application/x-tar'
	};
}
```

sindresorhus:
And don't change the existing fixture, but instead add another one.

sindresorhus:
@AshtonRoute Bump

Issue #248, opened 2019-10-15
Change proposal: break the fixed sample limit


Borewit:
### Limitation
Some file types cannot be determined reliable within the current 4k (or any reasonable single buffer limit). For example, an audio file with 600 kB ID3v2 header, may not be necessary an MP3 file. There are other audio formats which may be prefixed with an ID3v2 header.

### Proposed solution
Using a stream based solution, we could read information, starting from the beginning required to determine file type. This is an approach I use [music-metadata](https://github.com/borewit/music-metadata) using [strtok3](https://github.com/Borewit/strtok3) _tokenizer_, which supports _peeking_. It picks the most efficient way to jump to next required offset, depending if the underlying data is random accessible, like a file, blob or buffer. Or just read ahead in case of stream. This access abstraction can re-used from [readable-web-to-node-stream](https://github.com/Borewit/readable-web-to-node-stream).

> what happens if a file type can be determined by the first 4 bytes?

The first 4 bytes will be read, no more.

> what happens if the file is prefixed with ID3v2 header of 4 MB?

It will keep requesting data at the calculated offsets, in sequential manner, until it can determine the file type. But no more then required.

As this is major change, and will introduce a number of specialized function, e.g.:
```js
// Determine file type from a node.js file
const type1 = fileType.fromFile(file);

// Determine file type from a node.js stream
const type2 = fileType.fromStream(stream);

// Determine file type using a buffer as an input (where the buffer represents the entire file)
const type3 = fileType.fromBuffer(buffer)` 
```
A dedicated module for browser usage (using is effectively an adapter) offers browser specific functions:
```js
// Determine file type from a browser File or Blob
const type1 = browserFileType.fromBlob(blob);

// Determine file type from a browser ReadableStream
const type2 = browserFileType.fromStream(stream);
```
The browser part adapter is comparable with how I wrapped [music-metadata](https://github.com/borewit/music-metadata) with [music-metadata-browser](https://github.com/Borewit/music-metadata-browser).

I love to hear your feedback to make an assessment if this proposal is worth implementing.

Looking forward to hear your feedback.

Related issues: #210 

### Implementation status

Development branch: [stream-tokenizer](https://github.com/sindresorhus/file-type/tree/stream-tokenizer)

#### Progress / ToDo
- PR #261:
  - [x] Update AP
  - [x] Proof backward compatibility, being able to determine file type from a partial file as a buffer 
- PR #262:
[...]

PR #257, opened 2019-11-20
Unit tests rewritten with mocha and chai

Borewit:
I have difficulties debugging individual test cases in the current test setup.

Which may be just caused by the fact I am not very much used to the [ava test framework](https://github.com/avajs).

Although I have Linux machine around as well, I use a Windows machine for most open source development.

Debugging and fixing file-type is a pain for me. The development part is not Windows compliant.

I did a [change proposal](https://github.com/sindresorhus/file-type/issues/248), which involves drastic code change.

I would like to gain more control over the unit tests. And realize that this may not be the best answer for it, because applied the way of working I am used to.

There is 1 failure in the unit tests, which is interesting enough passing if executed in isolation. I have not been able to figure out yet what is going on.


sindresorhus:
Sorry, I don’t want to switch to Mocha.

Borewit:
Thanks for clarity, no problem.

PR #261, opened 2019-11-22
Initial integration with strtok3

Borewit:
Implementation of [*Change proposal: break the fixed sample limit* #248](https://github.com/sindresorhus/file-type/issues/248).

Initial integration with [strtok3](https://github.com/Borewit/strtok3), allowing the parser to read from the _tokenizer_.

The _tokenizer_ breaks the fixed buffer limitation, allowing as little, or as much data to be read to determine the file-type.

This PR:

1. Drafts the methods for node.js API.
2. Integrates with [strtok3](https://github.com/Borewit/strtok3)
3. Proof backwards compatibility: all unit tests passes

To be done:
1. Utilize the tokenizer in the `fromTokenizer()` method. (This will give the added value)
2. Update TypeScript definitions
3. Utilize the tokenizer instead of fixed buffer, in the _stream()_ method, this requires a kind of _teeing_ the input stream. I expect this to be very tricky.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 'use strict';
+const strtok3 = require('strtok3');
 const {
 	multiByteIndexOf,
 	stringToBytes,
@@ -12,17 +13,47 @@ const xpiZipFilename = stringToBytes('META-INF/mozilla.rsa');
 const oxmlContentTypes = stringToBytes('[Content_Types].xml');
 const oxmlRels = stringToBytes('_rels/.rels');
 
-const fileType = input => {
+async function fromStream(stream) {
+	const tokenizer = await strtok3.fromStream(stream);
+	try {
+		return await fromTokenizer(tokenizer);
+	} finally {
+		await tokenizer.close();
+	}
+}
+
+function fromBuffer(input) {
 	if (!(input instanceof Uint8Array || input instanceof ArrayBuffer || Buffer.isBuffer(input))) {
 		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\` or \`ArrayBuffer\`, got \`${typeof input}\``);
 	}
 
-	const buffer = input instanceof Uint8Array ? input : new Uint8Array(input);
+	const buffer = input instanceof Buffer ? input : Buffer.from(input);
 
 	if (!(buffer && buffer.length > 1)) {
 		return;
 	}
 
+	const tokenizer = strtok3.fromBuffer(buffer);
+	return fromTokenizer(tokenizer);
+}
+
+async function fromFile(path) {
+	const tokenizer = await strtok3.fromFile(path);
+	try {
+		return await fromTokenizer(tokenizer);
+	} finally {
+		await tokenizer.close();
+	}
+}
+
+async function fromTokenizer(tokenizer) {
+	const buffer = Buffer.alloc(tokenizer.fileSize);
+
+	await tokenizer.readBuffer(buffer, 0, tokenizer.fileSize);
+	return _fromBuffer(buffer);
+}
+
+function _fromBuffer(buffer) {
 	const check = (header, options) => {
 		options = {
 			offset: 0,
@@ -1034,22 +1065,19 @@ const fileType = input => {
 			mime: 'application/x-esri-shape'
 		};
 	}
-};
+}
 
-module.exports = fileType;
-
-Object.defineProperty(fileType, 'minimumBytes', {value: 4100});
-
-fileType.stream = readableStream => new Promise((resolve, reject) => {
+const stream = readableStream => new Promise((resolve, reject) => {
 	// Using `eval` to work around issues when bundling with Webpack
 	const stream = eval('require')('stream'); // eslint-disable-line no-eval
 
 	readableStream.on('error', reject);
-	readableStream.once('readable', () => {
+	readableStream.once('readable', async () => {
 		const pass = new stream.PassThrough();
-		const chunk = readableStream.read(module.exports.minimumBytes) || readableStream.read();
+		const chunk = readableStream.read(fileType.minimumBytes) || readableStream.read();
 		try {
-			pass.fileType = fileType(chunk);
+			const fileType = await fromBuffer(chunk);
+			pass.fileType = fileType;
 		} catch (error) {
 			reject(error);
 		}
@@ -1057,12 +1085,22 @@ fileType.stream = readableStream => new Promise((resolve, reject) => {
 		readableStream.unshift(chunk);
 
 		if (stream.pipeline) {
-			resolve(stream.pipeline(readableStream, pass, () => {}));
+			resolve(stream.pipeline(readableStream, pass, () => {
+			}));
 		} else {
 			resolve(readableStream.pipe(pass));
 		}
 	});
 });
+
+const fileType = {
+	fromStream,
+	fromTokenizer,
+	fromBuffer,
+	fromFile,
+	stream,
+	minimumBytes: 4100
+};
 
 Object.defineProperty(fileType, 'extensions', {
 	get() {
@@ -1075,3 +1113,5 @@ Object.defineProperty(fileType, 'mimeTypes', {
 		return new Set(supported.mimeTypes);
 	}
 });
+
+module.exports = fileType;
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "file-type",
-	"version": "12.3.1",
+	"version": "12.4.0",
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -12,17 +13,81 @@ const xpiZipFilename = stringToBytes('META-INF/mozilla.rsa');
...
+ * @returns {Promise<{ext, mime}|undefined>} Promise with file-type
+ * @public
+ */
+function fromBuffer(input) {
 	if (!(input instanceof Uint8Array || input instanceof ArrayBuffer || Buffer.isBuffer(input))) {
 		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\` or \`ArrayBuffer\`, got \`${typeof input}\``);
 	}
 
-	const buffer = input instanceof Uint8Array ? input : new Uint8Array(input);
+	const buffer = input instanceof Buffer ? input : Buffer.from(input);
```
sindresorhus: Note https://github.com/sindresorhus/file-type/commit/dc611b4c1293b7df9f9416d234e9b8935494519d

index.js
```diff
@@ -12,17 +13,81 @@ const xpiZipFilename = stringToBytes('META-INF/mozilla.rsa');
...
+		await tokenizer.close();
+	}
+}
+
+/**
+ * Determine file type from Buffer or Uint8Array.
+ * @param {Buffer | Uint8Array | ArrayBuffer} input - Partial file.
+ * @returns {Promise<{ext, mime}|undefined>} Promise with file-type
+ * @public
+ */
```
sindresorhus: I would prefer keeping the docs of the public methods to the TypeScript definition files.

index.js
```diff
@@ -1034,22 +1099,19 @@ const fileType = input => {
...
+}
 
-fileType.stream = readableStream => new Promise((resolve, reject) => {
+const stream = readableStream => new Promise((resolve, reject) => {
 	// Using `eval` to work around issues when bundling with Webpack
 	const stream = eval('require')('stream'); // eslint-disable-line no-eval
 
 	readableStream.on('error', reject);
-	readableStream.once('readable', () => {
+	readableStream.once('readable', async () => {
```
sindresorhus: You cannot just mark this as `async`. If anything throws it will cause an uncaught rejection. You need to wrap the whole body in a try/catch and then properly `reject` in the catch.

index.js
```diff
@@ -1064,14 +1126,13 @@ fileType.stream = readableStream => new Promise((resolve, reject) => {
...
-});
-
-Object.defineProperty(fileType, 'mimeTypes', {
-	get() {
-		return new Set(supported.mimeTypes);
-	}
-});
+module.exports = {
+	fromStream,
+	fromTokenizer,
```
sindresorhus: I don't think we should publicly expose the tokenizer.

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "file-type",
-	"version": "12.3.1",
+	"version": "12.4.0",
```
sindresorhus: Don't bump this.

readme.md
```diff
@@ -11,26 +11,51 @@ The file type is detected by checking the [magic number](https://en.wikipedia.or
...
 
+Determine file type from a file:
 ```js
-const readChunk = require('read-chunk');
-const fileType = require('file-type');
+import FileType from 'file-type';
+
+(async () => {
+    const fileType = await FileType.fromFile('/home/borewit/Pictures/background.png');
+    // fileType = {ext: 'png', mime: 'image/png'}
```
sindresorhus: Use tab-indentation.

index.js
```diff
@@ -1064,14 +1126,13 @@ fileType.stream = readableStream => new Promise((resolve, reject) => {
...
-	get() {
-		return new Set(supported.extensions);
-	}
-});
-
-Object.defineProperty(fileType, 'mimeTypes', {
-	get() {
-		return new Set(supported.mimeTypes);
-	}
-});
```
sindresorhus: These were using `Object.defineProperty` for a reason. They should not be changeable.

test.js
```diff
@@ -252,19 +266,19 @@ test('fileType.mimeTypes.has', t => {
...
 	});
 
 	t.notThrows(() => {
-		fileType(new Uint8Array());
+		fileType.fromBuffer(new Uint8Array());
 	});
 
 	t.notThrows(() => {
-		fileType(new ArrayBuffer());
+		fileType.fromBuffer(new ArrayBuffer());
```
sindresorhus:
```suggestion
		FileType.fromBuffer(new ArrayBuffer());
```

For consistency with the readme.

test.js
```diff
@@ -252,19 +266,19 @@ test('fileType.mimeTypes.has', t => {
 
 test('validate the input argument type', t => {
 	t.throws(() => {
-		fileType('x');
+		fileType.fromBuffer('x');
```
sindresorhus: This is now an async call, so this assertion is no longer correct. You need to use `t.throwsAsync`.

sindresorhus:
It definitely makes sense to make `fromBuffer` async, but I wonder if some users might have use-cases where they need a synchronous version too.

index.js
Borewit: The reason I am using `Buffer` is that I need pass a [Buffer](https://nodejs.org/api/buffer.html) to `strtok3.fromBuffer`.

It took me a while to figure out why UInt8Array instead of [Buffer](https://nodejs.org/api/buffer.html), please correct me if I am wrong, [browser compatibility](https://github.com/sindresorhus/file-type/pull/227#issuecomment-517741187) was the reason to use [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array).

To me it looks like the Buffer dependency in browser side code becomes less of a problem as module bundlers getting better in shimming this common dependency.

I am using the `buffer.copy` in [strtok3](https://github.com/Borewit/strtok3). Also [token-types](https://github.com/Borewit/token-types) is using Buffer specific functions, such as `readUInt16LE`, `readUInt16BE`, `readInt8` etc.

So [buffer](https://nodejs.org/api/buffer.html) functionality is actually being use. I could get rid of the Buffer dependency, in [strtok3](https://github.com/Borewit/strtok3) & [token-types](https://github.com/Borewit/token-types), but it means the functionality has to be duplicated using alternative ways. I am not sure if that is the way to go. As we scan files a bit deeper to determine the right file-type, I expect buffer's method will more and more used for decoding.

index.js
Borewit: Sure, so no method comments. I can also convert to TypeScript if you prefer,

index.js
Borewit: In either code, in case of reject, resolve will be called as well.,,

index.js
Borewit: Sorry, I am not familiar with that pattern, I will revert.

test.js
Borewit: Ah yes, missed that, thanks.

Borewit:
> It definitely makes sense to make `fromBuffer` async, but I wonder if some users might have use-cases where they need a synchronous version too.

The buffer inherits it's new async behavior from tokenizer. Is there a way to turn it into synchronous?

index.js
Borewit: I deliberately publicly exposed `fromTokenizer`. It allows to use specialized tokenizers like [tokenizer-http](https://github.com/Borewit/tokenizer-http) (HTTP-client using HTTP range requests). The same technology I used in [music-metadata-s3](https://github.com/Borewit/music-metadata-s3), to read very efficient from the Amazon S3 cloud. I am working on generic S3-tokenizer. 

Why not using a stream? This interface propagates the ignore calls, which can be converted in a very efficient requests of only the data really required, such as the first 384 bytes of the file, and the last 2k.

If you find it distracting, I will remove it, no problem. But it was exposed on purpose.

readme.md

package.json
Borewit: Sorry about that, reverted.

test.js
Borewit: Done

index.js
sindresorhus: Thanks for elaborating. It does make sense to expose it, but it needs to be documented what it's for.

index.js
sindresorhus: It's not a problem for me personally. It's just some browser users complained about size. But we can deal with it later if someone complains.

Borewit: I am sure few users won't like it.

index.js
Borewit: Done here: #279

sindresorhus:
> The buffer inherits it's new async behavior from tokenizer. Is there a way to turn it into synchronous?

Not that I'm aware of.

PR #262, opened 2019-11-28
Use tokenizer

Borewit:
Implementation of [*Change proposal: break the fixed sample limit* #248](https://github.com/sindresorhus/file-type/issues/248):

Migrating most file detection to [_tokenizer_](https://github.com/Borewit/strtok3#tokenizer)

Improved parsing using [_tokenizer_](https://github.com/Borewit/strtok3#tokenizer) of:
1. PNG parser (detecting APNG) demonstrates shows how much more elegant and deeper parser is possible if required
2. matroska sub-type selection
3. zip sub-type selection
3. ASF
3. MPEG / ID3
   * Remove fixture ID3 header starting at offset 1 (#101), this looks like to be based on a corrupt file.

As this involves significant effort, it would be good to hear your are happy with approach you would like me to take @sindresorhus.


index.d.ts
```diff
@@ -1,5 +1,5 @@
 /// <reference types="node"/>
-import {Readable as ReadableStream} from 'stream';
+import { Readable as ReadableStream } from 'stream';
```
sindresorhus: Don't do unrelated changes.

index.d.ts
```diff
@@ -238,70 +238,62 @@ declare namespace fileType {
...
 	type ReadableStreamWithFileType = ReadableStream & {
 		readonly fileType?: FileTypeResult;
 	};
-}
 
-declare const fileType: {
 	/**
-	Detect the file type of a `Buffer`/`Uint8Array`/`ArrayBuffer`. The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
+	Detect the file type of a `Buffer`/`Uint8Array`/`ArrayBuffer`.
+  The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
```
sindresorhus: Don't do unrelated changes.

package.json
```diff
@@ -13,6 +13,7 @@
 		"node": ">=8"
 	},
 	"scripts": {
+		"ava": "ava",
```
sindresorhus: You don't need this. Just run `npx ava` to run local binaries.

test.js
```diff
@@ -64,7 +64,6 @@ const names = {
 	],
 	mp3: [
 		'fixture',
-		'fixture-offset1-id3',
```
sindresorhus: Why was this fixture removed?

util.js
```diff
@@ -55,33 +42,15 @@ exports.tarHeaderChecksumMatches = buffer => { // Does not check if checksum fie
...
-	}
+exports.uint8ArrayUtf8ByteString = uint8ArrayUtf8ByteString;
 
-	return -1;
+/**
+ * ID3 UINT32 sync-safe tokenizer token
+ * 28 bits (representing up to 256MB) integer, the msb is 0 to avoid 'false syncsignals'
+ */
+exports.uint32SyncSafeToken = {
+	get: (buf, off) => {
```
sindresorhus:
```suggestion
	get: (buffer, offset) => {
```

util.js
```diff
@@ -55,33 +42,15 @@ exports.tarHeaderChecksumMatches = buffer => { // Does not check if checksum fie
...
+		return (buf[off + 3] & 0x7F) | ((buf[off + 2]) << 7) | ((buf[off + 1]) << 14) | ((buf[off]) << 21);
+	},
+	len: 4
```
sindresorhus:
```suggestion
	length: 4
```

index.d.ts
Borewit: The reason changed it, is because we are starting go beyond the "_magic numbers_". I will revert.

test.js
Borewit: I removed this fixture because I don't think fixture is a realistic representation of an MP3 file:
1. It is not normal an ID3 header starts at offset 1 or any other arbitrary offset in combination with an MPEG stream.
2. Common tools like Mp3Tag or Foobar2000, and a low level MPEG frame decoder I use, cannot make any sense of this file (neither from the MPEG frames).

I am not sure if it was this file, or another file from the same source, even the extension appeared to be wrong. This kind of fixtures are just causing confusion.

util.js
Borewit: Sorry, this part if the [token interface](https://github.com/Borewit/token-types/blob/de1957685596a1596df5daae4afaf89dbd709522/lib/index.ts#L28).
Cannot do that, will break the token.

sindresorhus: Alright. No worries. Although, I would recommend taking a look at https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prevent-abbreviations.md It really makes code more readable to not use abbreviations.

index.d.ts
```diff
@@ -243,65 +243,56 @@ declare namespace fileType {
...
-	readonly extensions: fileType.FileType;
+	 */
+	const extensions: fileType.FileType;
 
 	/**
 	Supported MIME types.
-	*/
-	readonly mimeTypes: fileType.MimeType;
+	 */
+	const mimeTypes: fileType.MimeType;
```
sindresorhus: Note: https://github.com/sindresorhus/file-type/commit/eff47afe8d0047c3e863f15af2372de1b3ff56b4

util.js
Borewit: I agree, despite the fact I cheat sometimes for good or bad reasons. I am firm believer in readable code. If that takes a bit more text that is fine. There are ways to optimize, if that is the reason to deviate from write verbose code. Easily being able to understand what is going on is one of them.

I did not design the token interface btw, I inherited it via [strtok2](https://www.npmjs.com/package/strtok2) (which I own and maintained and extended, but I did not write it initially). I wrote [strtok3](https://www.npmjs.com/package/strtok3) replacing the events with promises which allowed the _tokens_ to be typed. The _tokens_ remained intact.

Borewit:
Thanks a lot for merging @sindresorhus, allows to move on.

PR #263, opened 2019-11-30
Add Node.js script to analyze single file

Borewit:
Add a script helps during development to parse a single file or fixture.
```sh
cd file-type
./type.js fixture\fixture.m4a
```
outputs:
```
Parsing: fixture\fixture.m4a
MIME-type: audio/x-m4a
extension: m4a
```

This allows to focus on debugging a single file type.

### DIFF
--- a/type.js
+++ b/type.js
@@ -1,0 +1,20 @@
+#!/usr/bin/node
+const fs = require('fs');
+
+const fileType = require('.');
+
+if (process.argv.length !== 3) {
+	console.error('Expected path of the file to examine');
+	process.exit();
+}
+
+const file2test = process.argv[2];
+
+const type = fileType(fs.readFileSync(file2test));
+
+if (type) {
+	console.log(`MIME-type: ${type.mime}`);
+	console.log(`Extension: ${type.ext}`);
+} else {
+	console.log('Could not determine file type.');
+}


type.js
```diff
@@ -0,0 +1,21 @@
...
+const FileType = require('.');
+
+if (process.argv.length !== 3) {
+	console.error('Expected path of the file to examine');
+	process.exit();
+}
+
+const file2test = process.argv[2];
+
+console.log(`Parsing: ${file2test}`);
```
sindresorhus: This is moot as you already see it when you call the script.

type.js
```diff
@@ -0,0 +1,21 @@
...
+
+const fileType = FileType(fs.readFileSync(file2test));
+
+console.log(`MIME-type: ${fileType.mime}`);
+console.log(`extension: ${fileType.ext}`);
```
sindresorhus:
```suggestion
console.log(`Extension: ${fileType.ext}`);
```

type.js
```diff
@@ -0,0 +1,21 @@
...
+
+console.log(Buffer.from('M4A ').toString('hex'));
```
sindresorhus: ?

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+type.js
```
sindresorhus: Why is it ignored, but also included?

type.js
Borewit: Yeh, hmm, sorry, was some debug code.

.gitignore
Borewit: Confused it with .npmignore

Borewit:
Sorry, that was a bit of mess.

---
_off-topic:_


I remember why I added 

```json
"scripts": {
		"ava": "ava"
	}
```

That way I could debug the test suite easily from my IDE (& `--serial --verbose`), as far as I know, that does not work with `npx`.

sindresorhus:
> That way I could debug the test suite easily from my IDE (& --serial --verbose), as far as I know, that does not work with npx.

`npx ava --verbose` works fine.

sindresorhus:
Closing in favor of https://github.com/sindresorhus/file-type/pull/277.

PR #265, opened 2019-12-02
Separate Node.js dependency

Borewit:
Node.js functions are the default:

```js
const fileType = require('file-type');

(async () => {
    const fileType = await fileType.fromFile('/home/borewit/Pictures/background.png');
    // fileType = {ext: 'png', mime: 'image/png'}
})();
```

Common (internal) functions are accessible via sub-inclusion (`file-type/core`):

```js

(async () => {
    const fileType = await fileType.fromTokenizer(tokenizer);
    // fileType = {ext: 'png', mime: 'image/png'}
})();
```

This is part of the implementation of #248 _Change proposal: break the fixed sample limit_


sindresorhus:
I think it should be the other way around. `require('file-type')` should include the Node.js stuff, and `require('file-type/browser');` should be browser-only.

Borewit:
> I think it should be the other way around. require('file-type') should include the Node.js stuff, and require('file-type/browser'); should be browser-only.

Makes sense, I will adjust to your preference.

Borewit:
To convert the [Web-API readable stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader) (from browser part) into [Node.js readable stream](https://nodejs.org/api/stream.html#stream_readable_streams) (core) I planned to use [readable-web-to-node-stream](https://github.com/Borewit/readable-web-to-node-stream).

Would you like to keep `/browser` as part of `file-type` or put this in a dedicated module @sindresorhus?

sindresorhus:
> Would you like to keep /browser as part of file-type or put this in a dedicated module @sindresorhus?

I would prefer to keep everything in this package.

Borewit:
I will prepare a PR adding the browser functions within 2 days. 

If you have time,  it would be great if you can review & merge, these PR's to the [development branch](https://github.com/sindresorhus/file-type/tree/stream-tokenizer):
* #261
* #262 
* #265

core.d.ts
```diff
@@ -0,0 +1,319 @@
...
+
+	/**
+	Supported file extensions.
+	 */
+	const extensions: core.FileType;
+
+	/**
+	Supported MIME types.
+	 */
+	const mimeTypes: core.MimeType;
```
sindresorhus: Note: https://github.com/sindresorhus/file-type/commit/eff47afe8d0047c3e863f15af2372de1b3ff56b4

sindresorhus:
Merging this to unblock you, but don't forget to fix https://github.com/sindresorhus/file-type/pull/265#discussion_r360716604

PR #268, opened 2019-12-06
Use tokenizer browser

Borewit:
Part of implementation of #248

Add browser specific functions.

Get file type based on content from [Web API ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream):

```js
const FileType = require('file-type/browser');
(async () => {

    const response = await fetch("https://www.example.org/");
    const fileType = await FileType.fromStream(response.body);
})();

```

Get file type based on content from [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob):

```js
const FileType = require('file-type/browser');
(async () => {
    const fileType = await FileType.fromBlob(blob);
})();
```

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,0 +1,50 @@
+'use strict';
+const {ReadableWebToNodeStream} = require('readable-web-to-node-stream');
+const toBuffer = require('typedarray-to-buffer');
+const core = require('./core');
+
+async function fromStream(stream) {
+	const readableWebToNodeStream = new ReadableWebToNodeStream(stream);
+	const fileType = await core.fromStream(readableWebToNodeStream);
+	await readableWebToNodeStream.close();
+	return fileType;
+}
+
+async function fromBlob(blob) {
+	const buf = await convertBlobToBuffer(blob);
+	return core.fromBuffer(buf);
+}
+
+/**
+ * Convert Web API File to Node Buffer
+ * @param {Blob} blob Web API Blob
+ * @return {Promise<Buffer>}
+ */
+function convertBlobToBuffer(blob) {
+	return new Promise((resolve, reject) => {
+		const fileReader = new FileReader();
+		fileReader.addEventListener('loadend', event => {
+			let data = event.target.result;
+			if (data instanceof ArrayBuffer) {
+				data = toBuffer(new Uint8Array(event.target.result));
+			}
+
+			resolve(data);
+		});
+
+		fileReader.addEventListener('error', event => {
+			reject(new Error(event.message));
+		});
+
+		fileReader.addEventListener('abort', event => {
+			reject(new Error(event.type));
+		});
+
+		fileReader.readAsArrayBuffer(blob);
+	});
+}
+
+Object.assign(module.exports, core, {
+	fromStream,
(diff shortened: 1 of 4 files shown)


browser.js
```diff
@@ -0,0 +1,60 @@
...
+const toBuffer = require('typedarray-to-buffer');
+const core = require('./core');
+
+/**
+ * Determine file ReadableStream
+ * @param {ReadableStream} stream ReadableStream: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream
+ * @return {Promise<{ext, mime}|undefined>}  Promise with file-type
+ */
+async function fromStream(stream) {
+	const ns = new ReadableWebToNodeStream(stream);
```
sindresorhus: Don't use acronyms for variable names. Use full words.

browser.js
```diff
@@ -0,0 +1,60 @@
...
+		fileReader.onloadend = event => {
+			let data = event.target.result;
+			if (data instanceof ArrayBuffer) {
+				data = toBuffer(new Uint8Array(event.target.result));
+			}
+
+			resolve(data);
+		};
+
+		fileReader.addEventListener('error', error => {
```
sindresorhus:
```suggestion
		fileReader.addEventListener('error', event => {
```

browser.js
```diff
@@ -0,0 +1,60 @@
...
+			reject(new Error(error.type));
```
sindresorhus: Needs a better error than just the `type`.

browser.js
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Convert Web API File to Node Buffer
+ * @param {Blob} blob Web API Blob
+ * @return {Promise<Buffer>}
+ */
+function convertBlobToBuffer(blob) {
+	return new Promise((resolve, reject) => {
+		const fileReader = new FileReader();
+		fileReader.onloadend = event => {
```
sindresorhus: You can use `addEventListener` for this too.

Borewit:
Do you mind merging #261 & #262 first @sindresorhus? I am getting a little bit dizzy with all those outstanding PR's. Thanks a lot a for your detailed feedback, I know reviewing reviewing is lot of work.

sindresorhus:
Done :)

Borewit:
Okay, I cleaned up a bit.

I created to small PR with some fixing changes already merge (to prevent things getting mixed up):
* #274: Fixing some promise handling
* #275: Fixing README, `fileType` vs `FileType`

Than there is #265 preceding this PR, which move most functionality from `index.js` to `core.js`, which make `core.js` a polymorphic core (can be used by Node.js & browser). No rocket science, it essentially prevents a dependency on `fs` which is pretty Node.js specific. 

At last there is this PR, #268, which adds browser functions and corresponding dependencies. If you change your mind, as the change is more clear, and prefer moving to a separate module, this is a good escape point.
Yet another module to maintain, at the other hand it maybe easier to add browser unit tests and prevent browser specific dependencies to end up in the Node dependency tree.

sindresorhus:
This looks good too when you fix the merge conflict.

PR #270, opened 2019-12-18
Fix `extensions` and `mimeTypes` TypeScript types

armand1m:
Type definition for `extensions` and `mimeTypes` properties are wrong.

Considering they're a list of `T` (in this case, either `FileType` or `MimeType`), they should be defined as `ReadonlyArray<T>` instead of `T`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -296,7 +296,7 @@ declare const fileType: {
 	/**
 	Supported file extensions.
 	*/
-	readonly extensions: ReadonlyArray<fileType.FileType>;
+	readonly extensions: readonly fileType.FileType[];
 
 	/**
 	Supported MIME types.
--- a/index.d.ts
+++ b/index.d.ts
@@ -301,7 +301,7 @@ declare const fileType: {
 	/**
 	Supported MIME types.
 	*/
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -296,12 +296,12 @@ declare const fileType: {
 	/**
 	Supported file extensions.
 	*/
-	readonly extensions: fileType.FileType;
+	readonly extensions: ReadonlyArray<fileType.FileType>;
```
sindresorhus:
```suggestion
	readonly extensions: readonly fileType.FileType[];
```

index.d.ts
```diff
@@ -296,12 +296,12 @@ declare const fileType: {
...
 
 	/**
 	Supported MIME types.
 	*/
-	readonly mimeTypes: fileType.MimeType;
+	readonly mimeTypes: ReadonlyArray<fileType.MimeType>;
```
sindresorhus:
```suggestion
	readonly mimeTypes: readonly fileType.MimeType[];
```

armand1m:
probably will also have to change in the tests for the definitions as well

PR #273, opened 2019-12-20
Stop execution after `fileType.stream` error

Borewit:
In function `stream()`: prevent `resolve()` is called after `reject()`, in case of an error caught in `fileType()`.

Related comment: https://github.com/sindresorhus/file-type/pull/261/files#r360335513


sindresorhus:
> prevent resolve() is called after reject(), in case of an error caught in fileType().

A promise can only resolve once, with either success or failure, so that's moot. But I guess this makes it more readable.

PR #279, opened 2019-12-22
Document module function parseTokenizer 

Borewit:
* As requested [here](https://github.com/sindresorhus/file-type/pull/261#discussion_r360573288), documented `FileType.parseTokenizer()`.
* Updated API (method and parameters) documentation

### DIFF
--- a/core.d.ts
+++ b/core.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="node"/>
-import {Readable} from 'stream';
+import { Readable } from 'stream';
+import { ITokenizer } from 'strtok3/lib/type';
 
 declare namespace core {
 	type FileType =
@@ -260,23 +261,32 @@ declare namespace core {
  	If file access is available, it is recommended to use `fromFile()` instead
 
 	@param buffer - It works best if the buffer contains the entire file, it may work with a smaller portion as well
-	@returns The detected file type and MIME type or `undefined` when there was no match.
+	@returns The detected file type and MIME type, or `undefined` when there was no match.
 	 */
-	function fromBuffer(buffer: Buffer | Uint8Array | ArrayBuffer):  Promise<core.FileTypeResult | undefined>;
+	function fromBuffer(buffer: Buffer | Uint8Array | ArrayBuffer): Promise<core.FileTypeResult | undefined>;
 
 	/**
 	Detect the file type of a Node.js Readable.
   The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
 
 	@param stream - Node.js readable stream
-	@returns The detected file type and MIME type or `undefined` when there was no match.
+	@returns The detected file type and MIME type, or `undefined` when there was no match.
 	 */
 	function fromStream(stream: Readable): Promise<core.FileTypeResult | undefined>;
 
 	/**
+	 Detect the file type from an ITokenizer source.
+	 This method is used internally, but can also be used for a special 'tokenizer' reader.
+	 For more information see: https://github.com/Borewit/strtok3#tokenizer
+	 @param tokenizer - File source implementing the tokenizer interface.
+	 @returns The detected file type and MIME type, or `undefined` when there was no match.
+	 */
+	function fromTokenizer(tokenizer: ITokenizer): Promise<core.FileTypeResult>;
+
+	/**
 	Deprecated: The minimum amount of bytes needed to detect a file type. Currently, it's 4100 bytes, but it can change, so don't hard-code it.
 	 */
-  const minimumBytes: number;
+	const minimumBytes: number;
 
 	/**
 	Supported file extensions.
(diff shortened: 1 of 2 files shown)


core.d.ts
```diff
@@ -266,10 +267,19 @@ declare namespace core {
   The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
 
 	@param stream - Node.js readable stream
-	@returns The detected file type and MIME type or `undefined` when there was no match.
+	@returns The detected file type and MIME type via a promise, or `undefined` when there was no match.
 	 */
 	function fromStream(stream: Readable): Promise<core.FileTypeResult | undefined>;
 
+	/**
+	 * Parse audio from a ITokenizer source.
```
sindresorhus: Audio only? Why?

core.d.ts
```diff
@@ -257,7 +258,7 @@ declare namespace core {
  	If file access is available, it is recommended to use `fromFile()` instead
 
 	@param buffer - It works best if the buffer contains the entire file, it may work with a smaller portion as well
-	@returns The detected file type and MIME type or `undefined` when there was no match.
+	@returns The detected file type and MIME type via a promise, or `undefined` when there was no match.
```
sindresorhus: > via a promise

This is moot as the return type already says it's a promise.

readme.md
```diff
@@ -92,6 +92,33 @@ const fileType = require('file-type');
 })();
 ```
 
+Or from a _tokenizer_:
+
+A [_tokenizer_](https://github.com/Borewit/strtok3#tokenizer) propagates the internal read functions, 
+allowing alternative transport mechanisms, to access files, to be implemented and used.
+
+An example is [@tokenizer/http](https://github.com/Borewit/tokenizer-http) which requests data using range-requests.
+A difference with a conventional stream, is that it for example allows to only read the first 6, and the last 128 bytes.
```
sindresorhus: The docs in the TS file and readme should be in sync.

readme.md
```diff
@@ -92,6 +92,33 @@ const fileType = require('file-type');
...
+
+const audioTrackUrl = 'https://test-audio.netlify.com/Various%20Artists%20-%202009%20-%20netBloc%20Vol%2024_%20tiuqottigeloot%20%5BMP3-V2%5D/01%20-%20Diablo%20Swing%20Orchestra%20-%20Heroines.mp3';
+
+(async () => {
+  const httpTokenizer = HttpTokenizer.fromUrl(audioTrackUrl, config);
+  await httpTokenizer.init();
+
+  const fileType = await FileType.fromTokenizer(httpTokenizer);
+  console.log(fileType);
+  // { ext: 'mp3', mime: 'audio/mpeg' }
```
sindresorhus: Use tab-indentation.

Borewit: Oh yes, I cannot find an option in IntelliJ to override tab indents for markdown.

core.d.ts
Borewit: Copy past mistake

readme.md
Borewit: I missed the API documentation. I updated that path with the Typescript definitions.

core.d.ts
Borewit: But in the readme I kept the 

> _Returns a `Promise` resolving the _...

as the promise is not obvious

readme.md
Borewit: Fixed them one more time

readme.md
```diff
@@ -123,44 +154,83 @@ const url = 'https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg';
...
+
+If file access is available, it is recommended to use `fromFile()` instead
+
+| Parameter | Type                                                                                                                                                                                                                                                                            | Description                       |
+|-----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------|
+| path      | [`Buffer`](https://nodejs.org/api/buffer.html) or [`Uint8Array`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array)  or [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) | Buffer representing file content  |
```
sindresorhus: I don't want to use Markdown tables. They're annoying to maintain.

readme.md
```diff
@@ -90,6 +90,37 @@ const FileType = require('file-type');
 })();
 ```
 
+Or from a _tokenizer_:
```
sindresorhus: I think this should be moved to under the `.fromTokenizer()` docs as it's not something most would need.

readme.md
```diff
@@ -90,6 +90,37 @@ const FileType = require('file-type');
...
+
+A [_tokenizer_](https://github.com/Borewit/strtok3#tokenizer) propagates the internal read functions, 
+allowing alternative transport mechanisms, to access files, to be implemented and used.
+
+An example is [@tokenizer/http](https://github.com/Borewit/tokenizer-http),
+which requests data using [HTTP-range-requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests).
+A difference with a conventional stream and the [_tokenizer_](https://github.com/Borewit/strtok3#tokenizer),
+is that it is able to _ignore_ (seek, fast-forward) in the stream.
```
sindresorhus: Don't hard-wrap.

readme.md
```diff
@@ -123,44 +154,83 @@ const url = 'https://upload.wikimedia.org/wikipedia/en/a/a9/Example.jpg';
...
 ## API
 
-### fileType(input)
 
-Returns an `object` with:
+### FileType.fromBuffer(buffer)
+Detect the file type of a `Buffer`/`Uint8Array`/`ArrayBuffer`.
+The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
+
+If file access is available, it is recommended to use `fromFile()` instead
```
sindresorhus:
```suggestion
If file access is available, it is recommended to use `fromFile()` instead.
```

readme.md
Borewit: Converted to nested list

readme.md

PR #280, opened 2019-12-23
Tokenizer: fix merge conflicts with master branch

Borewit:
Merge with master to resolve merge conflicts.`

As the files have been renamed, merging is tricky.
* Manually added: https://github.com/sindresorhus/file-type/commit/57a5ad559438690ce4be246922eec7594f117f4d: Fix readme usage example
* #270 was already done via #278 
* #256 was changes to parser were already in, the fixture got now merged in as well


readme.md
```diff
@@ -54,18 +54,18 @@ const fs = require('fs');
 Or from a remote location:
 
 ```js
-const FileType = require('file-type');
-const http = require('http');
+const https = require('https');
+const fileType = require('file-type');
```
sindresorhus:
```suggestion
const FileType = require('file-type');
```

readme.md
```diff
@@ -54,18 +54,18 @@ const fs = require('fs');
...
-http.get(url, response => {
+https.get(url, response => {
 	response.on('readable', () => {
-		const chunk = response.read(FileType.minimumBytes);
+		const chunk = response.read(fileType.minimumBytes);
 		response.destroy();
 
-		console.log(FileType.fromBuffer(chunk)); // ToDo change to stream
-		//=> {ext: 'gif', mime: 'image/gif'}
+		console.log(fileType(chunk));
```
sindresorhus: This is the old API.

Borewit:
Updated and tested the Node.js examples.

I removed the browser _ToDo_'s with examples, but I have not tested those yet.

I hesitated to remove the `FileType.stream` example as it is limited to the 4k boundary and therefor no longer a recommended way to utilize file-type. Note that there are 2 streaming based examples already provided.

readme.md
Borewit: Note that I have updated the examples utilizing [node-fetch](https://www.npmjs.com/package/node-fetch).

It keeps the example clean, minimizes the difference between Node.js and browser.

readme.md

PR #281, opened 2019-12-27
Add CR3 File Type Support

LuttyYang:
Add CR3 File Type Support


Borewit:
Fixed conflicts and applied some changed in PR #281.

Please create a branch in the future, before submitting a PR. This helps makes merging on your end and on our end and reviewing a bit easier.

Issue #290, opened 2020-01-04
ASF/.asf detection is missing / broken


Borewit:
I believe there is no `.asf` detection, currently.

The [ASF](https://en.wikipedia.org/wiki/Advanced_Systems_Format) media container is now always mapped to `.wma` or `.wmv`, depending on the stream time.

If `.asf` is provided, it will be incorrectly detected as `.wma` or `.wmv`.

The way, I thinkt, it should work:
1. ASF using Windows Media Audio and Video codecs should map to `.wma` and `.wmv`.
1. Other ASF files should map to `.asf`.

Blocks #240.

ASF / `.wmv` video:
![image](https://user-images.githubusercontent.com/23255260/71770656-63a5cc80-2f2f-11ea-8667-68b2b1d280b4.png)

ASF / `.asf` video:
![image](https://user-images.githubusercontent.com/23255260/71770662-859f4f00-2f2f-11ea-858f-4045c6faa7c2.png)

Borewit:
I will add ASF codec support first in [music-metadata#392](https://github.com/Borewit/music-metadata/issues/392).

Borewit:
@sindresorhus, I think we should detect as `.asf` and drop support for `.wma` or `.wmv.`

Similar as:
- Matroska: #271, #272 
- MS-CFB: #377

I think we should detect the container format, and not guess the specific type if we are not sure.

sindresorhus:
> I think we should detect the container format, and not guess the specific type if we are not sure.

👍🏻

Issue #240, opened 2019-08-30
Add tests for `.asf`, `.ogm`, `.ogx`, and `.mpc` file types


sindresorhus:
https://github.com/sindresorhus/file-type/blob/a2692b3471030433f69c11e406456b32f3f49af5/test.js#L12-L17

* [x] `.asf` detection is not working: #290, failing fixture: PR #291
* [x] `ogm`, fixture added in PR #326
* [x] `ogx`, **artificial** fixture added in PR #327
* [x] `mpc`,  2 fixtures added in #176

Borewit:
`.asf` detection is not working: #290

PR #291, opened 2020-01-04
Add ASF unit test

Borewit:
1. Add ASF(https://en.wikipedia.org/wiki/Advanced_Systems_Format) / `.asf` file fixture.
2. Remove ignore `.asf` fixture in unit tests.

Currently failing do to: #290, related to #240

Fixture information:
- `fixture/fixture.asf ` source: https://samples.ffmpeg.org/asf-wmv/elephant.asf

image


sindresorhus:
We should have an array of `knownFailingTests` or something in the test file that makes the test use [`test.failing()`](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#failing-tests) instead of `test()`.

Borewit:
> We should have an array of `knownFailingTests` or something in the test file that makes the test use [`test.failing()`](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#failing-tests) instead of `test()`.

I could not agree more

Borewit:
Replaced by #311, as the source branch vanished.

PR #295, opened 2020-01-07
Add debug test run script

Borewit:
I know ava tests can be executed with `npx ava`, however, I cannot run such a target from my IntelliJ IDEA debugger. I don't want to put in and remove that small line by hand. 

I also ignored the IntelliJ project files. Some will say, I should put those in my user settings. However sharing such projects files can vary per project. I don't think it will harm anyone, and can be read add a project policy not to include these files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 node_modules
 yarn.lock
-.idea
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
 		"node": ">=10"
 	},
 	"scripts": {
+		"ava": "ava --serial --verbose",
 		"test": "xo && ava && tsd"
 	},
 	"files": [


sindresorhus:
> I also ignored the IntelliJ project files. Some will say, I should put those in my user settings. However sharing such projects files can vary per project. I don't think it will harm anyone, and can be read add a project policy not to include these files.

It should be in your [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c). I doesn't cause any harm, but it's not related to the project and it's a slippery slope to accepting any random files from a myriad of IDE's and tools.

sindresorhus:
Why do you need the `--serial --verbose` flags? There might be a better way if I know the context.

Borewit:
> There might be a better way if I know the context. How do they help with debugging?

Context:

`--serial` to run the unit test in sequential order. Helped me to fix and verify unit test one by one, when refactoring to [tokenizer](https://github.com/Borewit/strtok3#tokenizer).

`--verbose` to print out test results to the console, to give some visual feedback of test results.

Don't remember exactly, I had issues with debugging, and it seems that this combination made it easier.

I am used to [Mocha](https://mochajs.org/), [Jasmine](https://jasmine.github.io/), [junit5](https://junit.org/junit5/) from which I can easily pick out a group or individual test case to debug. 

I am using the [type.js](https://github.com/sindresorhus/file-type/blob/master/type.js), to parse individual files. However, that does not debug the test case itself, and it requires to setup each case individually. Running the `ava` script helps me to debug the test case itself.

PR #299, opened 2020-01-07
Fix TypeScript types in browser.d.ts

nwtgck:
Sorry for multiple PRs.
I fixed an issue of the type definition.

Before PR:
image

After PR:
image

---

I have a question. Is it better to add the following exports to browser.d.ts?

```ts
export type FileTypeResult = core.FileTypeResult;
export type FileExtension = core.FileExtension;
export type MimeType = core.MimeType;
```

---

- **One PR per file type.**
- Add the file type detection logic to the `index.js` file.
- Add the file extension to the `FileType` type in `index.d.ts`.
- Add the file's MIME type to the `MimeType` type in `index.d.ts`.
- Add the file extension to the `Supported file types` section in the readme, in the format ```- [`<extension>`](URL) - Format name```, for example, ```- [`png`](https://en.wikipedia.org/wiki/Portable_Network_Graphics) - Portable Network Graphics```

### DIFF
--- a/browser.d.ts
+++ b/browser.d.ts
@@ -1,4 +1,9 @@
+/// <reference lib="dom"/>
 import * as core from './core';
+
+export type FileTypeResult = core.FileTypeResult;
+export type FileExtension = core.FileExtension;
+export type MimeType = core.MimeType;
 
 /**
 Determine file type from a [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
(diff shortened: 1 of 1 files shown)


browser.d.ts
```diff
@@ -1,5 +1,10 @@
+/// <reference lib="dom"/>
 import * as core from './core';
 
+export type FileTypeResult = core.FileTypeResult;
```
Borewit: This can be written as well like:
```ts
export {FileTypeResult, FileExtension, MimeType} from './core';
```

sindresorhus: I thought so too, but I'm not 100% sure since these are TS types. TS has a lot of bugs, so I went with the simplest pragmatic solution.

Borewit: Sure no problem.

I should have been more careful writing the TypeScript.

PR #300, opened 2020-01-07
Make stream() method, stream based, by using teeing

Borewit:
**Request for comments, NOT SUITABLE FOR MERGING**

With version [v13.0.0](https://github.com/sindresorhus/file-type/releases/tag/v13.0.0) we moved away from the ~4 kilobyte sample boundary, allowing, if required, deeper analysis of files, streams or Buffers.

Very handy to determine the file type, before passing it to the corresponding parser.

The [`stream()` method](https://github.com/sindresorhus/file-type#stream), which aims to take a [`Readable` stream](https://nodejs.org/api/stream.html#stream_class_stream_readable), resolves the _file-type_, and returns, a `Readable` delivering the same content as received plus the resolved _file-type_, is still (or may) from this ~4 kilobyte limitation. Actually it may get a larger sample (chunk), but this is more luck then wisdom.

This PR is using a [stream tee module](https://github.com/Borewit/t-readable), splitting a [`Readable` stream](https://nodejs.org/api/stream.html#stream_class_stream_readable) into multiple [`Readable` streams](https://nodejs.org/api/stream.html#stream_class_stream_readable).

Handy isn't? One stream file for file type determination, one for the caller. Which probably would work great, if both streams would be consumed (read from) at the same time. Because we first read from the first copy of the stream, to an arbitrary point to determine the type, and only at that point we return the second copy.

Of-course it possible to cache all data read for determine the file type, and inject that in the returning stream. But that may cost an arbitrary amount of memory.

I thought is was worth sharing the issue and experience, even though I did not manage to crack the problem.

### DIFF
--- a/core.js
+++ b/core.js
@@ -1,6 +1,7 @@
 'use strict';
 const Token = require('token-types');
 const strtok3 = require('strtok3/lib/core');
+const {tee} = require('t-readable');
 const {
 	stringToBytes,
 	tarHeaderChecksumMatches,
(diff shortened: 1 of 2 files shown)


bencmbrook:
Hey @Borewit! Can you help me understand the current difference between `fromStream` and `stream`? (I built the `stream` method but was not involved in the byte length change for 13.0).

It sounds like the interface you're proposing is going to return a stream and a promise of the file-type of the stream?

Borewit:
@bencmbrook, thanks for your reply.

The return type is indeed different. `parseStream` is taken a `Readable` as an input, and returns the _file-type_, using a promise.

The `stream`, which you know better then I do, to my best understanding, passes the input stream to second + file-type as return type. I suppose this is designed to the ability to read from the input stream as if nothing happened, and be able to make decision based on the identified file type.

sindresorhus:
> The stream, which you know better then I do, to my best understanding, passes the input stream to second + file-type as return type. I suppose this is designed to the ability to read from the input stream as if nothing happened, and be able to make decision based on the identified file type.

Correct.

sindresorhus:
This does make sense. However, I think the `t-readable` package needs a lot more tests. Dealing with Node.js streams is very hard because they have a lot of surprising behavior. When it comes to cloning streams, you have to make sure error propagation is correctly handled. Back-pressure. Etc. Example: https://github.com/mcollina/cloneable-readable/blob/master/index.js

Borewit:
Thanks for your feedback @sindresorhus. At the moment I have no time to deep dive into this one, I get it out the way for the time being.

PR #301, opened 2020-01-07
Keep supporting Node v8

Borewit:
Restore Node.js version 8 compatibility.

I think the minimum Node engine version is set unnecessary high.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,4 @@ language: node_js
 node_js:
   - '12'
   - '10'
+  - '8'
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
 		"url": "sindresorhus.com"
 	},
 	"engines": {
-		"node": ">=10"
+		"node": ">=8"
 	},
 	"scripts": {
 		"test": "xo && ava && tsd"


sindresorhus:
Node.js 8 is EOL, so requiring Node.js 10 is totally fine.

https://twitter.com/nodekitten/status/1212114557939933184

Borewit:
That twitter statement is contradicting with the [description of the purpose of the _engines_ field](https://docs.npmjs.com/files/package.json#engines), it should indicate the required Node version, and not the desired version.

I don't think it is the responsibility of individual component to restrict applications to exclude older versions of Node, for any other reason then compatibility with the component itself. 

Consequences of lowering are:
1. Breaks compatibility of entire application 
2. Testing is no longer possible with Node version 8. 
3. Break compatibility with systems who are suffering from rare Node.js releases, like embedded systems. Synology for instance.

sindresorhus:
tl;dr This is how it works. Package authors drop support when the Node.js team does.

Every supported Node.js version comes with a certain burden, bugs/dependencies/etc. While the current `file-type` version works with Node.js 8, I'm not interested in guaranteeing that. The Node.js team is officially not supporting Node.js 8 anymore. It's unreasonable to expect module author to support a Node.js version that is not even supported by the Node.js project. Going forward, many dependencies will require Node.js 10, and I will have to require Node.js 10 then anyway to be able to upgrade. I did it now since we already had other breaking changes and needed to do a major release.

---

> Break compatibility with systems who are suffering from rare Node.js releases, like embedded systems. Synology for instance.

Synology is then effectively providing a Node.js version that is unsupported and will not receive security fixes. That sounds like much bigger issue than not being able to upgrade to the latest `file-type` version.

Borewit:
This module goes beyond scope of just Node.js, also browser engines and Electron are using it, and we deliberately choose that scope. This is a broader context than just Node.js. The engine, should indicate the engine requirement from an API and functionality point of view. The engine requirement is V8 and not V10.

At the point an issue is reported, which appears to be related to a Node.js 8, there is always the possibility to keep that issue unresolved, or increase the node engine requirement at that point. I understand the world is moving forward, and eventually all will follow. Yet I think it nonsense to exclude a compatible engine to anticipate on the fact that one of the implementations just run out maintenance.

I do run [music-metadata](https://github.com/Borewit/music-metadata) myself on a Synology. It is not very motivating contributing to a component breaking compatibility, just because of bloody digit in package file.

sindresorhus:
> The engine, should indicate the engine requirement from an API and functionality point of view.

No, the engine field indicates the supported version.

> or increase the node engine requirement at that point

Which would require yet another major version. Each major version is an inconvenience for thousands of users who have to manually upgrade and check what the breaking changes are.

> Yet I think it nonsense to exclude a compatible engine to anticipate on the fact that one of the implementations just run out maintenance.

And I think it's nonsense to have to support a version of Node.js that is not even supported by its creators.

However, seeing as you're a maintainer, I'm willing to keep Node.js 8 compatibility since you need it: https://github.com/sindresorhus/file-type/releases/tag/v13.0.3

Borewit:
> I'm willing to keep Node.js 8 compatibility

Thanks, but I would not have asked for it, if it would not believe this is the right thing to do or if it would cause any harm.

> NPM’s own docs say when defining engines:
> 
> “You can specify the version of node that your stuff works on” — NPM

[ref](https://itnext.io/node-engines-helping-developers-everywhere-avoid-phantom-bugs-2eef519604b2)

>engines: Use engines to specify specific versions of Node.js and/or npm that the project works with. It takes an object with a key for node and/or npm and a value that looks like what you’d have as values for dependencies that specifies a range of versions.

https://alligator.io/nodejs/package-json/

It is indeed not black and white. But the the way I read it, _specify the version  your stuff **works** with_, where one is free to narrow this down further, but that is the authors choice. 

Fact is that this an extremely dominant requirement. Every application deciding to upgrade will inherit this requirement. 453k candidates according to GitHub. If you take that in mind, it think it is wrong behavior of low-level component to set the scope smaller then strictly required.

From a support perspective, node engine V8 worked so far, there is no indication that this component would fall out of the sky tomorrow, using the same Node version 8 it always worked with.

But please keep updating your engines, I am not recommending anyone to stay behind.

PR #307, opened 2020-01-17
Prevent size related exceptions

### LINKED ISSUE #306: inconsistent behavior for small size buffer
I got an unexpected result from version 13.

version 12
```
const fileType = require('file-type');

console.log(fileType(Buffer.from('1234567890AB'))); // = undefined
console.log(fileType(Buffer.from('ABCD'))); // = undefined
```

version 13
```

(async () => {
  console.log(await FileType.fromBuffer(Buffer.from('1234567890AB'))); // = undefined
  console.log(await FileType.fromBuffer(Buffer.from('ABCD'))); // = throw error
})();
```

I hope this will be fixed.

Borewit:
Prevent exceptions:
* Parsing very small "file" sizes, fixes #306
* Parse files with unknown "file" size, like a stream required to fix failing unit tests #291.

### DIFF
--- a/core.js
+++ b/core.js
@@ -62,6 +62,10 @@ async function fromTokenizer(tokenizer) {
 	const check = (header, options) => _check(buffer, header, options);
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
+	if (!tokenizer.fileInfo.size) {
+		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;
+	}
+
 	await tokenizer.peekBuffer(buffer, 0, bytesRead, tokenizer.position, true);
 
 	// -- 2-byte signatures --
--- a/core.js
+++ b/core.js
@@ -63,7 +63,7 @@ async function fromTokenizer(tokenizer) {
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
 	if (!tokenizer.fileInfo.size) {
-		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;
+		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER; // if file-size is unknown, keep reading until EOF
 	}
 
 	await tokenizer.peekBuffer(buffer, 0, bytesRead, tokenizer.position, true);
--- a/core.js
+++ b/core.js
@@ -62,8 +62,9 @@ async function fromTokenizer(tokenizer) {
 	const check = (header, options) => _check(buffer, header, options);
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
+	// Keep reading until EOF if the file size is unknown.
 	if (!tokenizer.fileInfo.size) {
-		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER; // if file-size is unknown, keep reading until EOF
+		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;
 	}
 
 	await tokenizer.peekBuffer(buffer, 0, bytesRead, tokenizer.position, true);
--- a/core.js
+++ b/core.js
@@ -62,7 +62,7 @@ async function fromTokenizer(tokenizer) {
 	const check = (header, options) => _check(buffer, header, options);
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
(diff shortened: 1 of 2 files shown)


Borewit:
I am not really happy with way the `maybeless` flag is now provided to [`tokenizer.peekBuffer`](https://github.com/Borewit/strtok3/blob/master/README.md#method-tokenizerpeekbuffer) in:

```js
await tokenizer.peekBuffer(buffer, 0, bytesRead, tokenizer.position, true); //maybeless = true
```
I have created an issue: [strtok3#124](https://github.com/Borewit/strtok3/issues/124)

core.js
```diff
@@ -62,7 +62,11 @@ async function fromTokenizer(tokenizer) {
 	const check = (header, options) => _check(buffer, header, options);
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
-	await tokenizer.peekBuffer(buffer, 0, bytesRead);
+	if (!tokenizer.fileInfo.size) {
+		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;
+	}
```
sindresorhus: It's not clear to me why we need to do this. Could you add a short code comment?

test.js
```diff
@@ -459,9 +459,7 @@ test('odd file sizes', async t => {
 	const oddFileSizes = [1, 2, 3, 4, 5, 6, 7, 8, 10, 11, 12, 13, 255, 256, 257, 511, 512, 513];
 
 	for (const size of oddFileSizes) {
-		const buf = Buffer.alloc(size);
-		await t.notThrowsAsync(async () => {
-			await FileType.fromBuffer(buf);
-		}, `file size = ${size} bytes`);
+		const buffer = Buffer.alloc(size);
+		await t.notThrowsAsync(FileType.fromBuffer(buffer), `File size: ${size} bytes`);
```
sindresorhus: I have no idea why this is failing:

>  ✖ odd file sizes `t.notThrowsAsync()` must be called with a function or promise

`FileType.fromBuffer(buffer)` **is** a Promise.

Borewit: I am happy I am not the only one anymore who thinks the behavior of the async tests methods are weird. Passing a promise should be sufficient.

[test.js](https://github.com/sindresorhus/file-type/blob/master/test.js) is full of redundant `await` statements. 

I think in general async tests methods & await statements could be avoided by treating the test method acync if the return type of the test method is a promise.

sindresorhus: Ok, I know why it happens. When the file type cannot be detected, `.fromBuffer()` incorrectly returns `undefined` instead of `Promise<undefined>`.

core.js
Borewit: In some of the "level-2" parsers there using conditional based on the file-size:

https://github.com/sindresorhus/file-type/blob/e90035a1b46d3fe9347d90c8414ef9955c5f2701/core.js#L147

https://github.com/sindresorhus/file-type/blob/e90035a1b46d3fe9347d90c8414ef9955c5f2701/core.js#L225

https://github.com/sindresorhus/file-type/blob/e90035a1b46d3fe9347d90c8414ef9955c5f2701/core.js#L866

https://github.com/sindresorhus/file-type/blob/e90035a1b46d3fe9347d90c8414ef9955c5f2701/core.js#L927

In general they reading until the file-type detection completed, as long as the end of file was not reached.

If the file size is not known. By setting the `tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;` we keep reading until an end-of-file error is thrown.

test.js
Borewit: Aij

Borewit: It is missing `async` declaration:

https://github.com/sindresorhus/file-type/blob/e90035a1b46d3fe9347d90c8414ef9955c5f2701/core.js#L23

Borewit: I will create separate issue / PR for that. I am curious if the test cases are then working with a single await as well.

Borewit:
Force pushed a rebase on PR #310.

test.js
Borewit: And indeed, with the async fix, the async-test- methods do behave as expected :grin:

core.js
sindresorhus: Could you put something like your last paragraph in a code comment there?

PR #311, opened 2020-01-19
Add failing test for ASF

Borewit:
Replacement of PR #291, as I deleted the branch by deleting the forked repo

1. Add ASF(https://en.wikipedia.org/wiki/Advanced_Systems_Format) / `.asf` file fixture.
2. Remove ignore `.asf` fixture in unit tests.

Currently failing do to: #290, related to #240

Fixture information:
- `fixture/fixture.asf ` source: https://samples.ffmpeg.org/asf-wmv/elephant.asf

image


Borewit:
> We should have an array of `knownFailingTests` or something in the test file that makes the test use [`test.failing()`](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#failing-tests) instead of `test()`.

I am a bit stuck here. `fromBuffer` and `stream` are using the 4k limit (4 + 4096 bytes). I think that was a fair top keep the 4k limit in there for backward compatibility testing. Problem is, the the ASF fixture, it causes the _level 2_ parser to fall back on ASF. Which causes the file type to be guessed correctly. So the expected failure does not occur.

sindresorhus:
Comment it out or something for now maybe? But we cannot merge this when it fails the tests.

Borewit:
> Comment it out or something for now maybe? But we cannot merge this when it fails the tests.

Another option is to read the entire fixture into the buffer (instead of the 4k chunk). This will provide greater consistency between different parse methods. What do you think?

sindresorhus:
No, I don't think we should do that unless we drop the minimum bytes thing as we should test what users are using.

Borewit:
> No, I don't think we should do that unless we drop the minimum bytes thing as we should test what users are using.

I have no problem getting rid of _the minimum bytes thing_, it already became an arbitrary value. You can provide any chunk size you like, the smaller chunk, the less probability you will get with getting the right file type.

In general I would prefer to simplify the API, I would prefer not to expose possible file types (MIME-types and extensions) as enumerations and types neither. It's extra work to maintain, makes the module unnecessary heavy, causes unnecessary major API increments for each format added and I don't see a good reason for an application to depend on that. For internal testing it is very useful to have all possibilities listed.

Sorry for being a controversial here (once again).

sindresorhus:
> I have no problem getting rid of the minimum bytes thing, it already became an arbitrary value. You can provide any chunk size you like, the smaller chunk, the less probability you will get with getting the right file type.

Alright. Let's do it.

> I would prefer not to expose possible file types (MIME-types and extensions) as enumerations and types neither. It's extra work to maintain

We have to maintain them in the docs anyway. I don't see there being much extra overhead to maintain it in the code too. And there are some use-cases where it's useful. For example, checking if `file-type` supports a file format before passing a stream to it and if not, do your own custom checks or something else. They also don't add much byte-wise.

> causes unnecessary major API increments for each format added

They don't. Adding a file format is not a breaking change. Only removing one is, but that would be true even if we didn't expose MIME info.

PR #318, opened 2020-01-25
Update `strtok3` dependency

Borewit:
Update to [strtok3 v6](https://github.com/Borewit/strtok3/releases/tag/v6.0.0)

### DIFF
--- a/core.js
+++ b/core.js
@@ -77,7 +77,7 @@ async function _fromTokenizer(tokenizer) {
 		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;
 	}
 
-	await tokenizer.peekBuffer(buffer, 0, bytesRead, tokenizer.position, true);
+	await tokenizer.peekBuffer(buffer, {length: bytesRead, mayBeLess: true});
 
 	// -- 2-byte signatures --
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Much nicer API 👌

(I'm not going to do a new version yet as this change is not user-visible)

Borewit:
Thanks @sindresorhus. 

No rush in releasing a new version, after making the mistake updating and releasing [music-metadata@v6.3.1](https://github.com/Borewit/music-metadata/releases/tag/v6.3.1) with the same change, I realizes soon after it should have been a major version change instead of a patch.

The [`FileType.fromTokenizer(tokenizer)`](https://github.com/sindresorhus/file-type#filetypefromtokenizertokenizer) inherits the [tokenizer](https://github.com/Borewit/strtok3#tokenizer), which has changed. 
Probably no one is using this function, so non one would notice the difference, but still this a **major version change**.

Let's release it once the deprecation of [`FileType.minimumBytes`](https://github.com/sindresorhus/file-type#filetypeminimumbytes) is in.

sindresorhus:
> Let's release it once the deprecation of FileType.minimumBytes is in.

👍

PR #319, opened 2020-01-26
Deprecate minimumBytes

Borewit:
`FileType.minimumBytes` became obsolete, now we read beyond this 4k boundary. It is still possible to to use a sample of the first portion of the file. 

To ensure the best file identification, the entire file should be provided. Therefor it is highly recommended to use the `parseFile()`, which efficiently reads those, and only those, parts of the files, required for the file identification. 

Related: #248

### DIFF
--- a/browser.d.ts
+++ b/browser.d.ts
@@ -45,7 +45,6 @@ export declare function fromBlob(blob: Blob): Promise<core.FileTypeResult | unde
 
 export {
 	fromBuffer,
-	minimumBytes,
 	extensions,
 	mimeTypes
 } from './core';
--- a/core.d.ts
+++ b/core.d.ts
@@ -315,11 +315,6 @@ declare namespace core {
 	function fromTokenizer(tokenizer: ITokenizer): Promise<core.FileTypeResult | undefined>;
 
 	/**
-	Deprecated: The minimum amount of bytes needed to detect a file type. Currently, it's 4100 bytes, but it can change, so don't hard-code it.
-	*/
-	const minimumBytes: number;
-
-	/**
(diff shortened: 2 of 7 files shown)


jmac105:
Just to check my own understanding of this: there is still a minimum limit of 4100 bytes, but it now may be higher than that as well? I've seen a few cases in our implementation of this package where file type cannot be detected in an image because it is too small, this will still be an issue right?

Borewit:
> Just to check my own understanding of this: there is still a minimum limit of 4100 bytes, but it now may be higher than that as well? 

The minimum chunk size required to be able to detect a file type are 2 bytes. A significant number of file types will be detectable with the first 12 bytes. But some require 4 MB. 4100 bytes became on arbitrary chunk size, therefor this PR removed it from the API. 

Although not suitable for each use case, I highly recommend to use the method [`FileType.fromFile(filePath)`](https://github.com/sindresorhus/file-type#filetypefromfilefilepath) to determine the file type. That methods will query those parts of the file which are relevant to determine the file type.

> I've seen a few cases in our implementation of this package where file type cannot be detected in an image because it is too small, this will still be an issue right?

If you provide fewer bytes as required for the detection, the detection will fail or fall back to a more generic file type. This is not limited just to images.

jmac105:
Thanks. The files do not exist on disk as per current implementation but I'll have a look into that.

Borewit:
If you have the entire file in memory that is fine, in combination with `fromBuffer`. For an image file that is reasonable solution. Your browser has to load an image as well to be able to display it (and even decompress it).

PR #323, opened 2020-02-04
src: Add support for AI files (Adobe Illustrator)

### LINKED ISSUE #64: Support for Adobe Illustrator file ai
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/file-type/issues/64)

Hi,
i need to add the support for ai files. 
Thanks in advance.
Nick

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Submitted pull Requests
- [#323 src: Add support for AI files (Adobe Illustrator)](https://issuehunt.io/r/sindresorhus/file-type/pull/323)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/file-type/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

vladfrangu:
Thought I'd give a shot at adding support for Adobe's AI files. This might also allow other sequence-based searches to be done! Closes #64 (what an oldie!)

How does it work? There's no easy way to differentiate between AI and PDF. The first bytes are the same (`%PDF` in bytes). What you can look for however is the string `<xmp:CreatorTool>Adobe Illustrator 24.0 (Windows)</xmp:CreatorTool>` or similar, which is near the start of the file, before all the data!

The only other way I could think of is creating a PDF parser of sorts and reading it through that, but that doesn't sound easy enough 😝 

Link: https://en.wikipedia.org/wiki/Adobe_Illustrator_Artwork

### Question

If wanted, we could also just remove the limit when reading the buffer for a sequence check, which might help for stripped files.

#### Quick note

Some of the changes in this PR have been made to make xo happy.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#64: Support for Adobe Illustrator file ai](https://issuehunt.io/repos/19105307/issues/64)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

### DIFF
--- a/core.js
+++ b/core.js
@@ -1,6 +1,4 @@
 'use strict';
-const Token = require('token-types');
-const strtok3 = require('strtok3/lib/core');
 const {
 	stringToBytes,
 	tarHeaderChecksumMatches,
@@ -8,6 +6,8 @@ const {
 	uint8ArrayUtf8ByteString
 } = require('./util');
 const supported = require('./supported');
+const Token = require('token-types');
+const strtok3 = require('strtok3/lib/core');
 
 const minimumBytes = 4100;
 
@@ -56,6 +56,19 @@ function _check(buffer, headers, options) {
 	return true;
 }
 
+async function _checkSequence(sequence, tokenizer, options) {
+	options = {
+		position: 0,
+		...options
+	};
+
+	const buffer = Buffer.alloc(minimumBytes);
+
+	await tokenizer.peekBuffer(buffer, {position: options.position, length: 128, mayBeLess: true});
+
+	return buffer.includes(Buffer.from(sequence));
+}
+
 async function fromTokenizer(tokenizer) {
 	try {
 		return _fromTokenizer(tokenizer);
@@ -71,6 +84,7 @@ async function _fromTokenizer(tokenizer) {
 	const bytesRead = 12;
 	const check = (header, options) => _check(buffer, header, options);
 	const checkString = (header, options) => check(stringToBytes(header), options);
+	const checkSequence = (sequence, options) => _checkSequence(sequence, tokenizer, options);
 
 	// Keep reading until EOF if the file size is unknown.
 	if (!tokenizer.fileInfo.size) {
@@ -544,6 +558,14 @@ async function _fromTokenizer(tokenizer) {
 		return {
 			ext: 'wv',
 			mime: 'audio/wavpack'
+		};
+	}
+
+	const isAiFile = await checkSequence('Adobe Illustrator', {position: 1350});
+	if (isAiFile) {
+		return {
+			ext: 'ai',
+			mime: 'application/postscript'
 		};
 	}
 
--- a/core.js
+++ b/core.js
@@ -64,7 +64,7 @@ async function _checkSequence(sequence, tokenizer, options) {
 
 	const buffer = Buffer.alloc(minimumBytes);
 
-	await tokenizer.peekBuffer(buffer, {position: options.position, length: 128, mayBeLess: true});
+	await tokenizer.peekBuffer(buffer, {position: options.position, length: 512, mayBeLess: true});
 
 	return buffer.includes(Buffer.from(sequence));
 }
--- a/core.js
+++ b/core.js
@@ -561,7 +561,8 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
-	try {
+	if (checkString('%PDF')) {
+		// Check if this is an Adobe Illustrator file
 		const isAiFile = await checkSequence('Adobe Illustrator', {position: 1350});
 		if (isAiFile) {
 			return {
@@ -569,11 +570,8 @@ async function _fromTokenizer(tokenizer) {
 				mime: 'application/postscript'
 			};
 		}
-	} catch (_) {
-		// Ignore
-	}
 
-	if (checkString('%PDF')) {
+		// Assume this is just a normal PDF
 		return {
 			ext: 'pdf',
 			mime: 'application/pdf'
--- a/core.js
+++ b/core.js
@@ -56,15 +56,11 @@ function _check(buffer, headers, options) {
 	return true;
 }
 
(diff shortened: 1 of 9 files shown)


Borewit:
I need to fix an error message: Borewit/strtok3#147

core.js
```diff
@@ -56,6 +56,19 @@ function _check(buffer, headers, options) {
...
 
+async function _checkSequence(sequence, tokenizer, options) {
+	options = {
+		position: 0,
+		...options
+	};
+
+	const buffer = Buffer.alloc(minimumBytes);
+
+	await tokenizer.peekBuffer(buffer, {position: options.position, length: 512, mayBeLess: true});
```
Borewit: `position: options.position` is the default, redundant.

vladfrangu: It isn't, as when used, I overwrite position to `1350` for the Adobe Illustrator check

core.js
```diff
@@ -547,6 +561,18 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
+	try {
```
Borewit: If the file signature is the same as PDF, why don't you move this into `if PDF ...`

core.js
Borewit: `position: options.position` is the default, redundant.

vladfrangu:
@Borewit while you're here, hi!

I found an odd bug? (or misuse issue on my part) with `peekBuffer`, the tests for `aac` (`fixture-id3v2.aac` to be specific) always fail, and that happens only if the check for the AI files happens... 
How can `this.position` be changed with a `peekBuffer`? Or am I missing something?

core.js
Borewit: You can safely use `tokenizer.readBuffer` (advancing the position) once you detected the is PDF kind.

> What you can look for however is the string <xmp:CreatorTool>Adobe Illustrator 24.0 (Windows)</xmp:CreatorTool> or similar, which is near the start of the file, before all the data!

Do you think it possible to `tokenizer.read...` as long we read  _before all the data_ is met?

core.js
vladfrangu: That... Hasn't crossed my mind in fact, and would probably solve the error that was thrown by the peekBuffer function! Will move in my next commit

core.js
vladfrangu: Sorry, I'm a bit confused about what you mean by this 😅 
`options.position` can be changed depending if you pass in an options object to `_checkSequence`... Am I missing something?

Borewit:
> I found an odd bug? (or misuse issue on my part) with `peekBuffer`, the tests for `aac` (`fixture-id3v2.aac` to be specific) always fail, and that happens only if the check for the AI files happens...
> How can `this.position` be changed with a `peekBuffer`? Or am I missing something?

No, I think you have a point. Created issue: Borewit/strtok3#149

core.js
vladfrangu: I assume you're talking about `readToken`, correct? If so, we probably can read the buffer starting at the `1350`s byte, read 512 (for safety) and check if the resulting string includes `Adobe Illustrator`? Is that what you're referring to?

core.js
```diff
@@ -56,6 +56,19 @@ function _check(buffer, headers, options) {
 	return true;
 }
 
+async function _checkSequence(sequence, tokenizer, options) {
+	options = {
+		position: 0,
```
Borewit: 0 is a bad default position. Note ` that _check()` may be called in a recursion.

core.js
Borewit: Sorry I misread `position: tokenizer.position` is the default value. Better not to use absolute positions.

core.js
vladfrangu: Should I default the position to `tokenizer.position` then?

core.js
Borewit: I have not looked the format at all, but I was (naively) hoping there is a structured way of reading possible. Similar how the zip or png file is done.

vladfrangu: Here is a screenshot about how it looks. The items marked with a square are those that can (and will) change; specifically the document title (file name you save it as) and some arbitrary length... This is why the code skips `1350` bytes (which gives plenty leeway even for long file names) and reads `512` bytes which should *hopefully* catch the `CreatorTool`

image

core.js
Borewit: Yes, or `undefined`. Or use `await tokenizer.ignore(1350)`.

core.js
Borewit: I my god, [Adobe XMP](https://www.adobe.com/products/xmp.html) using [RDF](https://www.w3.org/RDF/), serialized with [RDF/XML](https://en.wikipedia.org/wiki/RDF/XML) inside PDF. And maybe I did not notice all layers yet.

core.js
vladfrangu: Done in https://github.com/sindresorhus/file-type/pull/323/commits/dc45850aa3aafe47eb41be288e46b54f78bb393b

core.js
vladfrangu: Solved in https://github.com/sindresorhus/file-type/pull/323/commits/dc45850aa3aafe47eb41be288e46b54f78bb393b

vladfrangu: If we can somehow figure out the length, then ignore till the title start, ignore title length, then till `CreatorTool`, we could probably use `readToken`... But, at least in this case... It's easier to read 512 bytes than do that. Plus, this method makes it work with other file types, if needed! Thoughts?

Borewit: If there is a header or field length encoded, it could be worth to try to decode that. Your current solution simple, which is also worth something.

PR #325, opened 2020-02-07
Remove the duplicated TIFF big-endian format check

patrickhulce:
I unfortunately don't have a file example and most TIFFs in the wild are little endian so it is fairly rare that this impacts things, but noticed it while perusing the code.

Introduced ~[1 month ago](https://github.com/sindresorhus/file-type/pull/292/files)


sindresorhus:
Can you add a code comment indicating which one is little endian and which one is big?

core.js
```diff
@@ -581,7 +581,7 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
-	if (check([0x49, 0x49, 0x2A, 0x0] || [0x4D, 0x4D, 0x0, 0x2A])) {
+	if (check([0x49, 0x49, 0x2A, 0x0]) || check([0x4D, 0x4D, 0x0, 0x2A])) {
```
patrickhulce:
```suggestion
        // TIFF-compatible formats can be encoded two ways:
        //  - little-endian/Intel (0x49)
        //  - big-endian/Motorola (0x4D)
	if (check([0x49, 0x49, 0x2A, 0x0]) || check([0x4D, 0x4D, 0x0, 0x2A])) {
```

patrickhulce:
> But please note that there is a duplicate big endian signature check here:

Ah I missed that thanks @Borewit! To my knowledge some of these formats handled here could potentially be big-endian as well, but agreed it's a bit messy, so just removing the confusing `||` for now makes sense to me. Thanks for the feedback!

core.js
```diff
@@ -581,7 +581,9 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
-	if (check([0x49, 0x49, 0x2A, 0x0] || [0x4D, 0x4D, 0x0, 0x2A])) {
+	// TIFF-compatible formats can be encoded two ways but little-endian (0x49) is most common.
+	// Big-endian TIFFs are handled below this block.
```
Borewit: This is little-endian I believe [ref](http://fileformats.archiveteam.org/wiki/TIFF#File_signatures).

Please add the `TIFF big-endian` comment to the other signature.

patrickhulce: Yep this is little-endian, sorry if the comment was confusing!

PR #331, opened 2020-02-12
Fix missing `fromTokenizer` type definition

mooyoul:
Currently `fromTokenizer` function only can be accessed by importing `file-type/core` due to missing type definition in type index.
This PR adds missing (re-exporting) `fromTokenizer` type definition.

----

BTW I don't want to leave off-topic message but I really LOVE that face-lifted API. Previously i manually peeked file in AWS S3 by sending ranged request. It added more complexity for inspecting file type. but now, we can just use handy tokenizers like `@tokenizers/s3` package!

@sindresorhus and @Borewit, Thanks for maintaining this project and Kudos for your work ❤️

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,6 +20,7 @@ export function fromFile(path: string): Promise<core.FileTypeResult | undefined>
 export {
 	fromBuffer,
 	fromStream,
+	fromTokenizer,
 	extensions,
 	mimeTypes,
 	stream


sindresorhus:
> BTW I don't want to leave off-topic message but I really LOVE that face-lifted API. Previously i manually peeked file in AWS S3 by sending ranged request. It added more complexity for inspecting file type. but now, we can just use handy tokenizers like @tokenizers/s3 package!

That would be a great tip for the readme ;)

Borewit:
> BTW I don't want to leave off-topic message but I really LOVE that face-lifted API. Previously i manually peeked file in AWS S3 by sending ranged request. It added more complexity for inspecting file type. but now, we can just use handy tokenizers like @tokenizers/s3 package!

@mooyoul very nice to hear someone is appreciating the API and someone is actually using the `fromTokenizer()` method. [@tokenizers/s3](https://github.com/Borewit/tokenizer-s3) was a bit of an idealistic project, you may be the first user directly using it. This _micro component_ is detached from [music-metadata-s3](https://github.com/Borewit/music-metadata-s3), to allow new [tokenizer consumers](https://github.com/Borewit/strtok3), like [file-type](https://github.com/sindresorhus/file-type), to utilize the same efficiency. 

Please star [@tokenizers/s3](https://github.com/Borewit/tokenizer-s3), and the dependent [@tokenizer/range](https://github.com/Borewit/tokenizer-range), to give me some confirmation these niche components are actually in use, and give newcomers some confirmation it works. 

> That would be a great tip for the readme ;)

I will describe the S3 use case with an small example in the readme.

PR #334, opened 2020-02-21
Fixes bug for short streams where stream ends on first read.

### LINKED ISSUE #330: stream.unshift() after end event
In some cases calling `readableStream.read` can reach the end of the stream (e.g. on a short file, or when stream emits everything in a single chunk), or stream may end for other reasons, causing `readableStream.unshift` to throw an error.

bencmbrook:
Fixes issue with calling `stream.unshift(chunk)` after a stream has already ended (that runtime error is shown here https://nodejs.org/api/stream.html#stream_readable_unshift_chunk_encoding).

Closes #330


bencmbrook:
I'd like to add a few more tests before merging this.

Borewit:
>Readable#unshift without the chunk arg is not well document and I'd like to make sure it works the way I think it does.

[documentation](https://nodejs.org/api/stream.html#stream_readable_unshift_chunk_encoding):

> Chunk of data to unshift onto the read queue. For streams not operating in object mode, chunk must be a string, Buffer, Uint8Array or null.

bencmbrook:
Alright, ready for review @Borewit / @sindresorhus 

I noticed the implementation was:
1. Read stream 1 and detect
2. Reset (unshift) stream
3. Attach pipe and return pipeline
 Read stream 1 and detect's actually better to:
1. Attach pipe
2. Read stream 1 and detect
3. Return pipeline

This solves the issue at its core, where stream 1 may end on read and make an unshift impossible.

PR #220, opened 2019-06-11
Add more ARW, DNG, NEF and TIF fixtures, and new detection patterns

mceachen:
Added a TIF and a bunch more RAW fixtures, truncated to 128 bytes, and new detection patterns. Fixes #219.

PR #337, opened 2020-03-10
Add current-generation Nikon D750 NEF example

mceachen:
(the added fixture was being incorrectly categorized as "image/tiff" before this PR)


Borewit:
Related PR: #220

PR #353, opened 2020-04-30
Add support for Encapsulated PostScript (EPS)

beaulac:
Adds support for [EPS files](https://en.wikipedia.org/wiki/Encapsulated_PostScript). Technically they are also PostScript, but they have an extra 32-byte header before the `%!` PostScript magic number, so they are not currently detected.

#### Detection
The [EPS spec](https://www.adobe.com/content/dam/acom/en/devnet/actionscript/articles/5002.EPSF_Spec.pdf) defines this extra header on page 23. The first 4 bytes must be `C5D0D3C6`.

#### Extension
EPS spec allows for several extensions (`.eps`, `.epsf`, `.epsi`, `.EPI`) depending on specific file contents beyond the scope of this library. The generic default `.eps` is used.

#### MIME type
EPS spec does not define a MIME type for EPS.
Several resources list the following possible types: 

* `application/postscript`
* `application/eps`
* `application/x-eps`
* `image/eps`
* `image/x-eps`

I chose `application/eps` to avoid confusion with "standard" postscript – happy to change it if desired.


bencmbrook:
Thanks for this PR @beaulac!

I believe we also need to check for signature bytes `25 21 50 53 2D 41 64 6F 62 65 2D 33 2E 30 20 45 50 53 46 2D 33 20 30`. This magic number, `C5D0D3C6` [seems to only apply to EPS files with a TIFF or WMF preview image](http://fileformats.archiveteam.org/wiki/Encapsulated_PostScript).

beaulac:
@bencmbrook Good point; I thought about including this, but this would be a breaking change as `25 21` (ASCII `%!`) is already the 'standard' postscript signature, so EPS files without a binary header are already detected by this library as ASCII PostScript `application/postscript` / `.ps`.

I can rearrange the checks so that the special case of `%!PS-Adobe-3.{0,1} EPSF-3.{0,1}` header is before the check for PostScript and that these are detected as EPS rather than PS, but this would change behaviour.

Thoughts?

bencmbrook:
Given it's becoming _more_ specific about the filetype, I think this would be a welcome change (with a minor version bump). @sindresorhus // @Borewit -- any input here?

sindresorhus:
> Given it's becoming more specific about the filetype, I think this would be a welcome change (with a minor version bump).

I agree

Borewit:
> Given it's becoming _more_ specific about the filetype, I think this would be a welcome change (with a minor version bump). @sindresorhus // @Borewit -- any input here?

Minor is fine with me, functionality has been added.

core.js
```diff
@@ -768,6 +778,13 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
+	if (check([0xC5, 0xD0, 0xD3, 0xC6])) {
```
bencmbrook, line 781: Please add the commented change request for other EPS types, and bump the minor version. Everything else looks great!

beaulac: The 'plaintext' EPS check is included in the PostScript check [here](https://github.com/sindresorhus/file-type/pull/353/files/1f121dcfeee45d02a4eef02ad62d8cf6d5b1907d#diff-04aff9558c8839f145fd5cc256639700R123) since they share the same first two bytes. The signature to check varies because of version numbers (`"%!PS-Adobe-X.X EPSF-X.X"` where each x is a single digit), so it only checks the fixed parts of the signature.

Borewit: > Please add the commented change request for other EPS types, and bump the minor version. Everything else looks great!

Better to leave the actual version increment to @sindresorhus.

sindresorhus: Yes, revert the version bump. You should generally never bump a version in a PR as it's up to the maintainers to decide when to release a new version and with what version (for example, they might be planning some breaking changes).

beaulac: @sindresorhus 👍 reverted

bencmbrook:
Thank you @beaulac!

PR #361, opened 2020-05-14
Add support for LZH

ia15076:
Add support for [LZH](https://en.wikipedia.org/wiki/LHA_(file_format)) archive files. 

**file type detection / magic bytes**

According to [the documentation of the header structure included with LHA for UNIX](https://github.com/jca02266/lha/blob/master/header.doc.jp), the LZH archive file describes the ID of the compression method in the third to seventh bytes. So I implemented it to detect based on that.

The possible IDs are as follows:

* `-lh0-`
* `-lh1-`
* `-lh2-`
* `-lh3-`
* `-lh4-`
* `-lh5-`
* `-lh6-`
* `-lh7-`
* `-lzs-`
* `-lz4-`
* `-lz5-`

Some unofficial IDs are not included in this document, but I didn't implement them.

In fact, `-lh5-` is the only one that is commonly used. If you don't want to increase the conditions too much, I think it's okay to remove the other detections.

**MIME type**

The MIME type that an LZH archive file can be are as follows:

* `application/x-lzh-compressed`
* `application/lha`
* `application/x-compress`
* `application/x-lha`
* `application/x-lha-compressed`
* `application/x-unknown-content-type-LhasaArchive`
* `application/x-zip-compressed`
* `binary/lzh`

I chose `application/x-lzh-compressed` because it is considered the most common from the description in [the Wikipedia article](https://en.wikipedia.org/wiki/LHA_(file_format)).

---

- [x] **One PR per file type.**
- [x] Add the file type detection logic to the `index.js` file.


core.js
```diff
@@ -1313,7 +1333,8 @@ const stream = readableStream => new Promise((resolve, reject) => {
 		const pass = new stream.PassThrough();
 		let outputStream;
 		if (stream.pipeline) {
-			outputStream = stream.pipeline(readableStream, pass, () => {});
+			outputStream = stream.pipeline(readableStream, pass, () => {
+			});
```
sindresorhus, line 1337: ?

ia15076: Oops! I will erase it...

Issue #336, opened 2020-02-26
Support PGP 


BrendanBall:
When I use the unix file command on a pgp encrypted file, it successfully detects it as `PGP RSA encrypted session key ...`
however using this package it returns `undefined`.

Is it possible to support pgp?

sindresorhus:
Sure. PR welcome :)

BrendanBall:
does this reimplement libmagic from scratch in pure javascript? I see there's another package `mmmagic` that just straight uses libmagic, it just unfortunately seems to be unmaintained. E.g. there's an old PR open to add a promise API.

Borewit:
> does this re-implement libmagic from scratch in pure javascript?

[file-type](https://github.com/sindresorhus/file-type) is written in pure JavaScript indeed.

Borewit:
Can you provide fixtures @BrendanBall?

BrendanBall:
sure

PR #367, opened 2020-05-29
Add support for PGP

akime-:
PGP description https://en.wikipedia.org/wiki/Pretty_Good_Privacy
mime description https://tools.ietf.org/html/rfc3156

FIxes https://github.com/sindresorhus/file-type/issues/336


core.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="node"/>
-import {Readable as ReadableStream} from 'stream';
-import {ITokenizer} from 'strtok3/lib/core';
+import { Readable as ReadableStream } from 'stream';
+import { ITokenizer } from 'strtok3/lib/core';
```
sindresorhus, line 3: Don’t do unrelated changes

akime-: Spaces removed

sindresorhus:
https://github.com/sindresorhus/file-type/pull/367/files#diff-1dd241c4cd3fd1dd89c570cee98b79ddR16-R17

package.json
```diff
@@ -204,4 +205,4 @@
 			"prefer-named-capture-group": "off"
 		}
 	}
-}
+}
```
sindresorhus, line 208: Unrelated change

akime-: Fixed

sindresorhus:
I commented https://github.com/sindresorhus/file-type/pull/367/files#issuecomment-636673451 for a reason. Please read it.

Issue #339, opened 2020-03-17
Word files (.docx) identified as application/zip


npalmius:
I know that similar issues (#312; #103; #66; #54) have been raised (and closed) before, but we've found some Word files (.docx) that aren't identified correctly using the latest version of file-type.

The test files are the three .docx files from https://file-examples.com/index.php/sample-documents-download/sample-doc-download/.

Note that version 12.4.2 works fine and detects the files as `application/vnd.openxmlformats-officedocument.wordprocessingml.document`, but versions 13 and 14 (including version 14.1.4)  incorrectly identify them as `application/zip`.

I appreciate that these are probably slightly special/unusual files, and I don't know how they were produced, but Word does open them correctly so I would say that they are valid Word documents.

For now we'll downgrade to version 12.4.2, but it would be great if this could be corrected in the current versions.

Borewit:
The algorithm to unzip docx has been changed after version 12.4.2. Instead of searching the buffer the file-header-signature, the offset to the next header is calculated. 

This has several advantages over the 12.4.2 algorithm:
- Able decode beyond 4k boundary
- More efficient, (compressed) file data is ignored 

The docx fixtures you provided are encoded with an unknown file length, which is currently not supported by the algorithm. That is the reason, the current file detection algorithm fails.

npalmius:
Thank you for looking in to this issue and for the description of the cause. It sounds like these files were produced using some obscure software (or maybe even manually). All I can really say is that Word opens them, so I think they are valid Word documents, but at the same time I appreciate that it's probably not a priority to support handling of a very unusual version of the file format. Of course, if it could be fixed then that would still be great.

As a side note, when investigating this issue on our side, the first thing we did was to use the file-type-cli tool to test the files, which correctly identified them because it still uses version 12.4.2 of file-type. This threw us off course a bit so it would be useful if the cli tool used the latest version of the library. On the other hand, then we wouldn't have realised that it worked correctly in 12.4.2!

Borewit:
> It sounds like these files were produced using some obscure software (or maybe even manually). 

So far, I have no reason to assume the provided samples are obscure.

I appropriate you reported it, and provided examples. That is good start. I am busy at the moment, I will investigate later what it takes to support the unknown-file-size flag.

npalmius:
That's a fair comment. Highly appreciate your time and effort on thie package!

tsmx:
[...]

PR #369, opened 2020-06-04
Fix ZIP header detection for MS Office files

hermandavid:
Hello, this PR fixes the issue #339.

For MS Office documents, some of the ZIP headers have set `compressedSize` to zero and that breaks the calculation of the next header offset. For such cases, this fix tries to find the next offset manually.


Borewit:
What about the buffer length. Do you think it is large enough to reliably detect the next header offset?

hermandavid:
I have added fixtures for MS Office files that are fixed by this PR.

Regarding the buffer length, I'm honestly not sure if Office produces incomplete ZIP headers that would be big enough to exceed the current buffer size. To make it more robust, we could scroll through the buffer until we either find a next ZIP header or we reach the end of file.

Borewit:
> I have added fixtures for MS Office files that are fixed by this PR.

Very nice.

> Regarding the buffer length, I'm honestly not sure if Office produces files with incomplete ZIP headers that would be big enough to exceed the current buffer size. To make it more robust, we could scroll through the buffer until we either find a next ZIP header or we reach the end of file.

It is also worth something to keep it simple, let's see how this works out first.

PR #370, opened 2020-06-07
Fix error when using `FileType.fromStream()` on small files

### LINKED ISSUE #341: Very small file (11 byte)
Expected - return undefined 
Bug - got exception  'End-Of-Stream'
Scenario - Sent very small file (11 bytes), the file data sent as stream.

`let fileType = await FileType.fromStream(passThroughStream)`

Borewit:
- Update [strtok3](https://github.com/Borewit/strtok) to [6.0.1](https://github.com/Borewit/strtok3/releases/tag/v6.0.1), fixes #341
- Add unit test revealing issue #341: test parsing short files in combination with [`FileType.fromStream`](https://github.com/sindresorhus/file-type#filetypefromstreamstream)

Note that the actual fix is implemented via a patch release in [strtok3 version 6.0.1](https://github.com/Borewit/strtok3/releases/tag/v6.0.1).

### DIFF
--- a/test.js
+++ b/test.js
@@ -1,6 +1,6 @@
 import path from 'path';
 import fs from 'fs';
-import {Readable} from 'stream';
+import stream from 'stream';
 import test from 'ava';
 import {readableNoopStream} from 'noop-stream';
 import FileType from '.';
@@ -286,7 +286,7 @@ test('.stream() method - empty stream', async t => {
 
 test('.stream() method - short stream', async t => {
 	const bufferA = Buffer.from([0, 1, 0, 1]);
-	class MyStream extends Readable {
+	class MyStream extends stream.Readable {
 		_read() {
 			this.push(bufferA);
 			this.push(null);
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -1,6 +1,6 @@
 import path from 'path';
 import fs from 'fs';
-import stream from 'stream';
+import {Readable} from 'stream';
```
sindresorhus, line 3: `Readable` what? Named imports only makes sense when it's clear enough from the name what it is.

```suggestion
import {Readable as ReadableStream} from 'stream';
```

test.js
```diff
@@ -484,11 +484,28 @@ test('validate the repo has all extensions and mimes in sync', t => {
...
 
+class BufferedStream extends Readable {
+	constructor(buffer) {
+		super();
+		this.push(buffer);
+		this.push(null);
+	}
+
+	_read() {
+	}
```
sindresorhus, line 495:
```suggestion
	_read() {}
```

sindresorhus:
Can you give the PR a descriptive title?

Issue #375, opened 2020-06-24
Add `asar` support


Richienb:
[asar](https://github.com/electron/asar#format) is an archive format that was primarily developed to enclose Electron applications. Here's a [zip of a `fixture.asar`](https://github.com/sindresorhus/file-type/files/4823187/fixture.zip) that can be used in the unit tests.

// @sindresorhus

sindresorhus:
Can you link to a source that describes the actual format in technical terms?

Richienb:
@sindresorhus https://github.com/electron/asar#format

dyoshikawa:
I want to contribute about this.
What mime type is asar?

Richienb:
@dyoshikawa `application/asar`

dyoshikawa:
I made PR https://github.com/sindresorhus/file-type/pull/378 for this issue.

PR #378, opened 2020-07-14
Add support for Asar

dyoshikawa:
Fixes https://github.com/sindresorhus/file-type/issues/375

## The official page of the file format
https://github.com/electron/asar#format


core.js
```diff
@@ -1355,6 +1355,23 @@ async function _fromTokenizer(tokenizer) {
 			}
 		}
 	}
+
+	// Check for Asar
```
Richienb, line 1359: This can be simplified down to:

```js
	// Check for Asar
	try {
		const headerSize = buffer.readUInt32LE(12);
		const header = buffer.slice(16, headerSize + 16).toString();
		JSON.parse(header);
		return {
			ext: 'asar',
			mime: 'application/asar'
		};
	} catch (_) {}
```

dyoshikawa: Thanks. Merged your suggestion.

Richienb:
// @sindresorhus

Richienb:
@dyoshikawa In the pull request message, change `For` to `Fixes` so that the issue can be automatically closed.

dyoshikawa:
@Richienb Thanks. Changed my message.

core.js
Borewit: If would prefer to first make a assessment (an educated guess) on binary level if a file is a JSON / Pickle file before parsing it as JSON. To prevent unnecessary parsing of JSON and prevent errors.

Borewit: In case this sample is Asar file, is the header indeed always exact 16 bytes?

dyoshikawa: Thanks for review.
I'm sorry, I don't have knowledge to detect type as binary level.

> In case this sample is Asar file, is the header indeed always exact 16 bytes?

Worked in multiple Asar files when I tried.
But, I didn't find an official evidence for this thing.

Do anyone have an idea?

Borewit: To me it looks like [asar](https://github.com/electron/asar#format) is using [Pickle](https://chromium.googlesource.com/chromium/src/+/master/base/pickle.h) to serialize.

> Worked in multiple Asar files when I tried.
> But, I didn't find an official evidence for this thing.

There must be some clue this supposed to work. If you cannot find a written explanation, you could try to derive it from the source code.: [pickle.h](https://chromium.googlesource.com/chromium/src/+/master/base/pickle.h), [pickle.cc](https://chromium.googlesource.com/chromium/src/+/master/base/pickle.cc). To me it looks like Pickle is starting with header portion of 4 bytes (32 bit).

If the length is indeed fixed for Asar (I don't think this is the case for any Pickle format), you may use the length withing the 4 byte portion as an initial signature. The `ext` in the JSON header part may be another one.

dyoshikawa: Thanks your advices.
I am investigating. 
It still takes a while.

dyoshikawa: @Borewit 

I found clues from `electron/asar` source code and wrote comments about this.
I also added a checking binary process, but this method may be poor.
Could you review again?

CC: @Richienb

core.js
```diff
@@ -1355,6 +1355,24 @@ async function _fromTokenizer(tokenizer) {
...
+	try {
+		// Check header_size
+		buffer.readUInt32LE(4);
+
+		// Check header
+		// Header's first char should be "{" because expect JSON
+		if (check([0x7B], {offset: 16})) {
```
Richienb, line 1366:
```suggestion
		if (checkString("{", {offset: 16})) {
```

core.js
```diff
@@ -1355,6 +1355,24 @@ async function _fromTokenizer(tokenizer) {
...
 	}
+
+	// Check for Asar
+	try {
+		// Check header_size
+		buffer.readUInt32LE(4);
+
+		// Check header
+		// Header's first char should be "{" because expect JSON
+		if (checkString("{", {offset: 16})) {
```
dyoshikawa, line 1366: Possibly may be `{"files": {`.

Richienb:
As it turns out, asar doesn't actually have a mime type so it should be set to `application/octet-stream`. https://github.com/electron/asar/issues/199#issuecomment-660589582

dyoshikawa:
> As it turns out, asar doesn't actually have a mime type so it should be set to application/octet-stream. electron/asar#199 (comment)

Can we use original mime type name?
ex. `application/x-asar`

core.js
dyoshikawa: Thx.
I fixed double quote to single.

Richienb:
@dyoshikawa Not sure. Try asking that on https://github.com/electron/asar/issues/199

dyoshikawa:
@Richienb 

https://www.iana.org/assignments/media-types/media-types.xhtml

Unregisterd mime types in this library seems to already exist.
ex. `application/x-sqlite3`, `application/x-google-chrome-extension` and more.

https://tools.ietf.org/html/rfc6838#section-3.4

> Subtype names with "x." as the first facet may be used for types intended exclusively for use in private, local environments.

So `application/x.asar` looks good to me.

Richienb:
🤷‍♂️

core.js
Borewit: I will, I am sorry I let you wait.

bencmbrook:
> Unregisterd mime types in this library seems to already exist.

These unregistered types are still used elsewhere so there's at least _some_ traction behind them.

> So application/x.asar looks good to me.

Since `x.` is **"intended exclusively for use in private, local environments"**, is doesn't work here.

I don't think this library should create new MIME types. It should be the responsibility of the community around the format to decide on and register their MIME types. Given it's `application/x-asar` is not used elsewhere, I'd like to see this MIME as `application/octet-stream`.

Borewit:
Consistent with existing code, I think we can use `application/x-asar`. `/x-` indicates an unregistered MIME-type.
We use an extensions and MIME-types to express the identified file type.
Both of them are not perfect. Wherever possible we shall comply with the standard.

core.js
```diff
@@ -1355,6 +1355,24 @@ async function _fromTokenizer(tokenizer) {
 			}
 		}
 	}
+
+	// Check for Asar
+	try {
+		// Check header_size
+		buffer.readUInt32LE(4);
```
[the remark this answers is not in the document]
Borewit: So we determine asar / pickle in 2 steps. 
1. First a quick scan on Pickle heuristics
2. After passing 1, we can spend all resources to ensure it is the right file type, as only very few files will be tested

Please put the try catch, as it is expensive, in part 2.

If you know the header size is first byte make use of it! Does it have a fixed value, minimum value or a likely maximum value.

dyoshikawa:
@Borewit Thanks.
I rewrote checking logic by 2 steps.
And, I adopted `application/x-asar` as mime type.

package.json
```diff
@@ -178,7 +178,8 @@
 		"avif",
 		"eps",
 		"lzh",
-		"pgp"
+		"pgp",
+	  	"asar"
```
sindresorhus, line 182: Incorrect indentation

readme.md
```diff
@@ -402,6 +402,7 @@ Returns a set of supported MIME types.
 - [`eps`](https://en.wikipedia.org/wiki/Encapsulated_PostScript) - Encapsulated PostScript
 - [`lzh`](https://en.wikipedia.org/wiki/LHA_(file_format)) - LZH archive
 - [`pgp`](https://en.wikipedia.org/wiki/Pretty_Good_Privacy) - Pretty Good Privacy
+- [`asar`](https://github.com/electron/asar#format) - Asar
```
sindresorhus, line 405: Give it a better description

package.json
Richienb:
```suggestion
		"asar"
```

readme.md
Richienb:
```suggestion
- [`asar`](https://github.com/electron/asar#format) - Archive format primarily used to enclose Electron applications
```

core.js
```diff
@@ -1355,6 +1355,24 @@ async function _fromTokenizer(tokenizer) {
 			}
 		}
 	}
+
+	// Check for Asar
+	if (buffer.length > 16 && checkString('{"files":{', {offset: 16})) {
+		try {
+			// Check header_size
+			buffer.readUInt32LE(4);
```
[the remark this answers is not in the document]
dyoshikawa: Thanks.
I think this line should delete because your opinion.
Do you think so, too?

core.js
```diff
@@ -1355,6 +1355,24 @@ async function _fromTokenizer(tokenizer) {
 			}
 		}
 	}
+
+	// Check for Asar
+	if (buffer.length > 16 && checkString('{"files":{', {offset: 16})) {
```
Richienb, line 1360:
```suggestion
	if (checkString('{"files":{', {offset: 16})) {
```

core.js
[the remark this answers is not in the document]
Richienb: Go ahead.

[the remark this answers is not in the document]
dyoshikawa: Yes, sir.

core.js
dyoshikawa: Thx.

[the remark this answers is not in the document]
dyoshikawa: Thx. I adopted it.

core.js
```diff
@@ -1367,7 +1385,8 @@ const stream = readableStream => new Promise((resolve, reject) => {
 		const pass = new stream.PassThrough();
 		let outputStream;
 		if (stream.pipeline) {
-			outputStream = stream.pipeline(readableStream, pass, () => {});
+			outputStream = stream.pipeline(readableStream, pass, () => {
```
dyoshikawa, line 1388: @Borewit 
Is it needed?
This diff looks only a new line.

Richienb: Best remove this newline.

core.js
```diff
@@ -1195,6 +1195,24 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
+	if (check([0x04, 0x00, 0x00, 0x00]) && buffer.length >= 16) { // Rough & quick check Pickle/ASAR
```
Richienb, line 1198: Are all of these comments necessary?

Richienb:
@sindresorhus

dyoshikawa:
Thank you for everyone!

PR #380, opened 2020-07-22
Fix the `.extensions` TypeScript type

ethanresnick:


sindresorhus:
You need to update index.test-d.ts too.

Issue #352, opened 2020-04-30
Tar detection seems broken


jongross:
### Tar detection on MacOs seems broken

OS: MacOs 10.15.4
hardware: iMac 5k

node version: 13.13.0
file-type version: 14.2.0

I have built a couple different tarballs on various different systems, and none of them are being detected as tar files.

I've stepped through the ID process, and the headers (masked as your code indicates) are just plain old not matching, so it seems to be failing in the header matches, not in the checksum.

Below is a hexdump snippet from the two tar files that aren't being matched:

`hexdump -n 256 pest.tar
0000000 70 72 6f 63 5f 6d 6f 64 75 6c 65 2e 78 6d 6c 00
0000010 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
0000060 00 00 00 00 31 30 30 36 34 34 20 00 20 20 20 20
0000070 20 30 20 00 20 20 20 20 20 30 20 00 20 20 20 20
0000080 37 32 32 32 36 37 37 20 31 31 36 34 31 31 37 30
0000090 33 34 34 20 20 31 32 33 35 37 20 00 30 00 00 00
00000a0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
0000100`

` hexdump -n 256 orig.tar
0000000 2e 2f 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0000010 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
0000060 00 00 00 00 30 30 30 37 35 35 20 00 30 30 30 37
0000070 36 35 20 00 30 30 30 30 32 34 20 00 30 30 30 30
0000080 30 30 30 30 30 30 30 20 31 33 36 35 32 36 31 33
0000090 31 35 31 20 30 31 31 33 31 36 00 20 35 00 00 00
00000a0 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
*
0000100`

I was curious if you could help me identify why the tar match is failing for me.

sindresorhus:
Is the `file` command able to identify them?

jongross:
> Is the `file` command able to identify them?

Yes, the `file` command knows what they are.

Borewit:
Can you please provide a sample file @jongross? Zip the tar archive to be able to drop it here. Preferably very small file size.

Borewit:
Related #241

jongross:
[...]

PR #384, opened 2020-08-02
Replace `uint8ArrayUtf8ByteString()` with `Buffer.toString()`.

Borewit:
Working on #352, I thought we could get rid of  `uint8ArrayUtf8ByteString()` and rely on Node Buffer string conversions instead.

This PR replaces  `uint8ArrayUtf8ByteString()` with [`Buffer.toString()`](https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end).

Fine-tuned the decoding of:
* TAR header `chksum` to [ASCII](https://en.wikipedia.org/wiki/ASCII)
* MPEG-4 brand major identifier to [ISO 8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1)

### DIFF
--- a/util.js
+++ b/util.js
@@ -2,10 +2,6 @@
 
 exports.stringToBytes = string => [...string].map(character => character.charCodeAt(0));
 
-const uint8ArrayUtf8ByteString = (array, start, end) => {
-	return String.fromCharCode(...array.slice(start, end));
-};
-
 exports.tarHeaderChecksumMatches = buffer => { // Does not check if checksum field characters are valid
 	if (buffer.length < 512) { // `tar` header size, cannot compute checksum without it
 		return false;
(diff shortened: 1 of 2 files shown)


util.js
```diff
@@ -30,7 +26,7 @@ exports.tarHeaderChecksumMatches = buffer => { // Does not check if checksum fie
 		signedBitSum += byte & MASK_8TH_BIT; // Add signed bit to signed bit sum
 	}
 
-	const readSum = parseInt(uint8ArrayUtf8ByteString(buffer, 148, 154), 8); // Read sum in header
+	const readSum = parseInt(buffer.toString('binary', 148, 154), 8); // Read sum in header
```
bencmbrook, line 29: Curious why the encoding parameter is binary here? At first glance this seems like it's not a 1:1 replacement

bencmbrook:
Looks good to me — @Borewit any anticipated impacts on Browser support? My understanding is file-type already depends on Buffer, which means Browser users will already have to [polyfill for Buffer](https://github.com/feross/buffer) regardless of this change

util.js
Borewit: Correct, it is not. I don't think UTF is the proper encoding: [_"All characters in header blocks are represented by using 8-bit characters in the local variant of ASCII"_](https://www.gnu.org/software/tar/manual/html_node/Standard.html). Having another look to it, neither `binary` (this is an alias for `ISO8859-1`), I will set it to `ascii`. It should not make a difference at the end, in case it is a tar file, only text encoded digits are expected, which are all encoded the same way in these encoding types.

Borewit:
> any anticipated impacts on browser support?

Should be good. I use the same function in [music-metadata-browser](https://github.com/Borewit/music-metadata-browser), using the same [polyfill for Buffer ](https://github.com/feross/buffer).

Borewit:
Thanks for reviewing @bencmbrook!

PR #385, opened 2020-08-04
Fix tar detection

### LINKED ISSUE #241: Check archive types before any other type

### LINKED ISSUE #352: Tar detection seems broken

Borewit:
Fixes #352
Fixes #241 

Addresses issues:
* ~~Tar header was checked at a point only 256 of 512 required bytes (TAR header size) had been read.~~ Duplicate tar test, one running to early.
* Initial tar file was not identifying positive on the sample provided with issue #352.

Changes:
* Removed duplicate tar check, one at the point where 512 have been read yet. 
* Fixed initial tar signature check, allow space padding in checksum field. Fixes #352
* Updated the sum value decoding in line with [tar for Node](https://savannah.gnu.org/projects/tar)  
* Add truncated (first 512 bytes) fixture (`fixture-spaces.tar`) of sample file provided in #352.

References:
* [GNU tar project](https://savannah.gnu.org/projects/tar)
* [tar for Node](https://github.com/npm/node-tar)
* ~~[Tar signature](https://www.filesignatures.net/index.php?page=search&search=TAR&mode=EXT)~~

ToDo:
* [x] Fix tar detection
* [x] Add unit test which fails on check `check([0x30, 0x30, 0x30, 0x30, 0x30, 0x30], {offset: 148, mask: [0xF8, 0xF8, 0xF8, 0xF8, 0xF8, 0xF8]})`, by having `0x20` (space) at byte offset 148.


util.js
```diff
@@ -2,14 +2,20 @@
 
 exports.stringToBytes = string => [...string].map(character => character.charCodeAt(0));
 
+function readTarSum(buffer) {
+	return parseInt(buffer.toString('utf8', 148, 154).replace(/\0.*$/, '').trim(), 8); // Read sum in header
+}
+
+exports.readTarSum = readTarSum;
```
bencmbrook, line 9: Does this need to be a separate util?

Can `tarHeaderCheckumMatches` simply fail early if `!isNaN(readTarSum(buffer)`?

Borewit: Good point.

Borewit:
Thanks for reviewing @bencmbrook!

PR #386, opened 2020-08-07
Simplify `tar` checksum calculation

Borewit:
Simplify function `util.tarHeaderChecksumMatches`:
- Decoding based on [node-tar](https://github.com/npm/node-tar/blob/8b605d7def4cb15370b573ea45a1831a2fe1cd9d/lib/header.js#L97) (note that they do not include the `signedBitSum`)
- Add comments
- Preserve (TAR header) offset as an optional second parameter

This PR should not change any behavior.

Related to #385

### DIFF
--- a/util.js
+++ b/util.js
@@ -3,11 +3,12 @@
 exports.stringToBytes = string => [...string].map(character => character.charCodeAt(0));
 
 /**
- * Return true if the tar checksum is valid
- * @param {Buffer} buffer holding tar header [offset ... offset + 512]
- * @param {number} off TAR header offset
- * @returns {boolean} rue if the tar checksum is valid, otherwise false
- */
+Checks whether the TAR checksum is valid.
+
+@param {Buffer} buffer - The TAR header `[offset ... offset + 512]`.
+@param {number} offset - TAR header offset.
+@returns {boolean} `true` if the TAR checksum is valid, otherwise `false`.
+*/
 exports.tarHeaderChecksumMatches = (buffer, off) => {
 	off = off ? off : 0;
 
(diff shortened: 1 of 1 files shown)


util.js
```diff
@@ -2,43 +2,32 @@
...
-		return false;
-	}
+/**
+ * Return true if the tar checksum is valid
+ * @param {Buffer} buffer holding tar header [offset ... offset + 512]
+ * @param {number} off TAR header offset
+ * @returns {boolean} rue if the tar checksum is valid, otherwise false
+ */
+exports.tarHeaderChecksumMatches = (buffer, off) => {
+	off = off ? off : 0;
```
sindresorhus, line 13: `off` => `offset`

util.js
```diff
@@ -2,43 +2,32 @@
...
-	if (buffer.length < 512) { // `tar` header size, cannot compute checksum without it
-		return false;
-	}
+/**
+ * Return true if the tar checksum is valid
+ * @param {Buffer} buffer holding tar header [offset ... offset + 512]
+ * @param {number} off TAR header offset
+ * @returns {boolean} rue if the tar checksum is valid, otherwise false
+ */
+exports.tarHeaderChecksumMatches = (buffer, off) => {
```
sindresorhus, line 12:
```suggestion
exports.tarHeaderChecksumMatches = (buffer, offset = 0) => {
```

Borewit: I thought this was only possible in Typescript, good to know!.

Borewit:
Thanks for reviewing @sindresorhus & @bencmbrook

PR #388, opened 2020-08-24
Replace MSI detection with Microsoft CFB

### LINKED ISSUE #377: Document expected return values of unsupported file types
Because these following file types are not supported, it would be helpful to users to know what ext/mime `file-type` will return for them (eg: `application/x-msi`)

### LINKED ISSUE #387: False positives test are broken
Following Microsoft documents are **not** tested as [`.msi` (Windows Installer)](https://en.wikipedia.org/wiki/Windows_Installer) or [[MS-CFB]: Compound File Binary File Format](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cfb/):
https://github.com/sindresorhus/file-type/blob/4a4d39cd905844625d6dd8540ecae3a870d6dd08/test.js#L178-L182

```js
const path = require('path');

(async () => {
  const fileType = await FileType.fromFile(path.join('fixture', 'fixture-xls.msi'));
  console.log(fileType);
})();
```
Outputs:
```
undefined
```

Related:
* 702f795e4c4baa4bc32e8eecc5423f11b71fbbc3: Reduce msi false-positives
* #162 False-positives for the `msi` detection

Borewit:
Replace MSI detections with [Microsoft CFB](https://en.wikipedia.org/wiki/Compound_File_Binary_Format).

Fix #377: [Microsoft CFB](https://en.wikipedia.org/wiki/Compound_File_Binary_Format) maps to `.msi`. `.msi` is CFB based, but some other file types as well.

[Microsoft CFB](https://en.wikipedia.org/wiki/Compound_File_Binary_Format) is identified as:
```js
{
  ext: cfb,
  mime: 'application/x-msi'
}
```

Related: #387, the present fixtures did not test positive on [Microsoft CFB](https://en.wikipedia.org/wiki/Compound_File_Binary_Format).

- [x] Need to understand #387 first.

Fix #387, remove false false positive tests.


niftylettuce:
You can use this sample doc for CFB: https://github.com/spamscanner/spamscanner/blob/master/test/fixtures/sample.doc

niftylettuce:
@sindresorhus can you release a new version to npm? many thanks as always ❤️ 🙇 👏 🎉 

thank you @Borewit

sindresorhus:
https://github.com/sindresorhus/file-type/releases/tag/v15.0.0

PR #399, opened 2020-09-20
Add the ability to register third-party extensions

### LINKED ISSUE #340: Extending file-type
Dear maintainers. 
First, I really appreciate that file-type has been keep maintaining. Thank you for your efforts & contributions.

## Background

Recently, file-type was refactored to utilize @Borewit's awesome strtok3 packages.
With this changes, we can navigate binary stream easily & effectively.
Now, file-type uses only one Tokenizer interface regardless of source. All public methods creates Tokenizer internally and pass created tokenizer to file-type matcher.

Beside these refactors, developers cannot easily extend file-type.
If someone needs to add new file type support, There are only two choices:

1. Fork file-type package and customize
Pros: It does not occur additional i/o if chunk was cached internally
Cons: Less maintainability

2. Run custom file-type matchers if file-type returns `undefined`.
Pros: More maintainability
Cons: It occurs additional I/O

As you can see, There are no effective way to extend file-type.

### Expose Matcher interface

So, It would be nice if we support additional matchers.

Let's suppose there is a matcher like this:

```typescript
interface MatchedType {
  ext: string;
  mime: string;
}

// write custom matcher
async function webpMatcher(tokenizer: ITokenizer): Promise<MatchedType | void>  {
  await tokenizer.ignore(8);
  const buf = Buffer.alloc(4);
  await tokenizer.peekBuffer(buf);
  if (buffer.compare(Buffer.from([0x57, 0x45, 0x42, 0x50])) === 0) {
    return { ext: "webp", mime: "image/webp" };
  }
}
```

Then, Provide custom matchers like these ways:

```typescript
// quick & dirty - provide custom matchers as options
fileType.fromFile(path, { matchers: [webpMatcher] });

// or provide factroy like axios does
const fileType = FileType.create({ matchers: [webpMatcher] });

// or provide instance methods to add matchers to specific position (head or tail)
const fileType = FileType.create();
fileType.appendMatcher(webpMatcher);
fileType.prependMatcher(webpMatcher);

```

> I'm not sure which way is cleanest. Please leave your feedback!

## Future

With this idea, Developers can easily extend file-type.

From file-type's perspective, file-type can have multiple 3rd-party type matchers which can be released to npm independently, or managed via Github organization/Scoped packages (e.g. `@file-type/core` for 
built-it matchers, `@file-type/image` for image file type matchers)

Also it can be used solving unsupported features like https://github.com/sindresorhus/file-type/issues/264#issuecomment-568439196. 
For example, Developers can write custom TSV/CSV type matchers and publish to npm even if file-type does not provide built-in TSV/CSV file type matchers.

Borewit:
### FileType.addDetection(detection)

Register a detection function. This function fill be called after the file-type, and other detections completed.
the `detection(tokenizer, fileType)` method has two arguments:
* [tokenizer](https://github.com/Borewit/strtok3#tokenizer)
* `fileType`, which is `undefined` or the return value of a detected file type.

Resolves #340.

Issues:
- Not possible to overrule existing file-type detection

### DIFF
--- a/core.d.ts
+++ b/core.d.ts
@@ -370,7 +370,7 @@ declare namespace core {
 	*/
 	function stream(readableStream: ReadableStream): Promise<core.ReadableStreamWithFileType>
 
-	type Detection = (tokenizer: ITokenizer, fileType? : core.FileTypeResult) => Promise<core.FileTypeResult | undefined>;
+	type Detection = (tokenizer: ITokenizer, fileType?: core.FileTypeResult) => Promise<core.FileTypeResult | undefined>;
 
 	function addDetection(detection: Detection): void;
 }
--- a/core.d.ts
+++ b/core.d.ts
@@ -369,6 +369,10 @@ declare namespace core {
 	```
 	*/
 	function stream(readableStream: ReadableStream): Promise<core.ReadableStreamWithFileType>
+
+	type Detection = (tokenizer: ITokenizer, fileType? : core.FileTypeResult) => Promise<core.FileTypeResult | undefined>;
+
+	function addDetection(detection: Detection): void;
 }
 
 export = core;
--- a/readme.md
+++ b/readme.md
@@ -270,12 +270,13 @@ Returns a set of supported MIME types.
 
 ### FileType.addDetection(detection)
 
-Register a detection function. This function fill be called after the file-type, and other detections completed.
-the `detection(tokenizer, fileType)` method has two arguments:
-* [tokenizer](https://github.com/Borewit/strtok3#tokenizer)
-* `fileType`, which is `undefined` or the return value of a detected file type.
+Register a detection function. This function will be called after `file-type` and other detections complete.
 
-If the detector returns `undefined` the tokenizer.position should be 0. That allows other detectors to parse the file. 
+The `detection(tokenizer, fileType)` method has two arguments:
+- [tokenizer](https://github.com/Borewit/strtok3#tokenizer)
+- `fileType`, which is `undefined` or the return value of a detected file type.
+
+If the detector returns `undefined`, the `tokenizer.position` should be 0. That allows other detectors to parse the file. 
 
 ## Supported file types
 
--- a/readme.md
+++ b/readme.md
@@ -268,6 +268,15 @@ Returns a set of supported file extensions.
 
 Returns a set of supported MIME types.
 
+### FileType.addDetection(detection)
+
+Register a detection function. This function fill be called after the file-type, and other detections completed.
+the `detection(tokenizer, fileType)` method has two arguments:
+* [tokenizer](https://github.com/Borewit/strtok3#tokenizer)
+* `fileType`, which is `undefined` or the return value of a detected file type.
+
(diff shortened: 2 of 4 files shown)


readme.md
```diff
@@ -268,6 +268,16 @@ Returns a set of supported file extensions.
 
 Returns a set of supported MIME types.
 
+### FileType.addDetection(detection)
+
+Register a detection function. This function will be called after `file-type` and other detections complete.
```
sindresorhus, line 273: What other detections?

readme.md
```diff
@@ -268,6 +268,16 @@ Returns a set of supported file extensions.
...
+
+The `detection(tokenizer, fileType)` method has two arguments:
+- [tokenizer](https://github.com/Borewit/strtok3#tokenizer)
+- `fileType`, which is `undefined` or the return value of a detected file type.
+
+If the detector returns `undefined`, the `tokenizer.position` should be 0. That allows other detectors to parse the file.
```
sindresorhus, line 279: Why cannot `file-type` reset the position automatically?

core.d.ts
```diff
@@ -369,6 +369,10 @@ declare namespace core {
 	```
 	*/
 	function stream(readableStream: ReadableStream): Promise<core.ReadableStreamWithFileType>
+
+	type Detection = (tokenizer: ITokenizer, fileType? : core.FileTypeResult) => Promise<core.FileTypeResult | undefined>;
+
+	function addDetection(detection: Detection): void;
```
sindresorhus, line 375: This needs the same docs as the readme.

sindresorhus (changes requested):

readme.md
Borewit: Because underlying source can be a stream. A stream can only be read once.

sindresorhus: Then it should be made clear that `the `tokenizer.position` should be 0` should not be done if it's a stream.

readme.md
jonyyz: Proprietary software file extensions that are not well known

Ugzuzg: Or even plain text files that the users will be willing to support. Say, there's not enough confidence to be able to detect an svg or HTML to include it in the core, but the users could accept that and provide their own detection which will be enough for their use case.

sindresorhus: Yes, that's the whole point of this PR. That's not what I'm asking. I'm saying that "other detections" is not clear enough in the docs.

jonyyz: I think "other" is probably intended to leave it open for extension.  Examples could be provided but are probably not going to be exhaustive.  Perhaps what you're asking for is some examples?

Borewit:
Sorry I am a bit unresponsive this time.  I very busy at the moment. Thanks for all your feedback.

sindresorhus:
Gentle bump in case you have time and just forgot about this. If you're still busy, feel free to ignore.

sindresorhus:
I'm going to close this as it has been open for a very long time without any activity. A new PR can be opened if when/if you get back to this ;)

PR #415, opened 2020-11-20
Add support for 3mf

Ugzuzg:
It follows the Open Packaging Conventions, so the detection is very similar to docx or odt.

> The 3D Model part name SHOULD contain two segments, the first being "/3D/" and the second with the extension ".model" on the last segment, for example "/3D/3dModel.model".
> -- https://github.com/3MFConsortium/spec_core/blob/master/3MF%20Core%20Specification.md#22-part-naming-recommendations

https://www.iana.org/assignments/media-types/model/3mf


core.js
```diff
@@ -318,6 +318,13 @@ async function _fromTokenizer(tokenizer) {
 					};
 				}
 
+				if (zipHeader.filename.startsWith('3D/') && zipHeader.filename.endsWith('.model')) {
```
sindresorhus, line 321:
```suggestion
				if (zipHeader.filename.startsWith('/3D/') && zipHeader.filename.endsWith('.model')) {
```

Ugzuzg: There's no leading slash in the filenames of any of the zip-based formats. The above will fail.
xlsx has the same implementation: https://github.com/sindresorhus/file-type/blob/master/core.js#L314-L319

sindresorhus (approved):

Borewit (approved):
Looks good to me. Another way, to determine the file, would be search for `[Content_Types].xml` , unzip that file, parse the XML, use the content-types. But not feasible in this context, I think this is the best way indeed.

Issue #422, opened 2020-12-18
Support Microsoft Compiled HTML Help


Richienb:
> [**Microsoft Compiled HTML Help**](https://en.wikipedia.org/wiki/Microsoft_Compiled_HTML_Help) is a Microsoft proprietary online help format, consisting of a collection of HTML pages, an index and other navigation tools. 

Filename extension: `chm`
Media type: [`application/vnd.ms-htmlhelp`](https://www.iana.org/assignments/media-types/application/vnd.ms-htmlhelp)

The magic number is the text [`ITSF`](https://en.wikipedia.org/wiki/Microsoft_Compiled_HTML_Help#File_format) in ASCII.

sindresorhus:
PR welcome

PR #424, opened 2020-12-18
Add support for CHM

Richienb:
> [**Microsoft Compiled HTML Help**](https://en.wikipedia.org/wiki/Microsoft_Compiled_HTML_Help) is a Microsoft proprietary online help format, consisting of a collection of HTML pages, an index and other navigation tools. 

Filename extension: `chm`
Media type: [`application/vnd.ms-htmlhelp`](https://www.iana.org/assignments/media-types/application/vnd.ms-htmlhelp)

The magic number is the text [`ITSF`](https://en.wikipedia.org/wiki/Microsoft_Compiled_HTML_Help#File_format) in ASCII.

Fixes: #422


sindresorhus (approved):

core.js
```diff
@@ -1326,6 +1326,13 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
+	if (checkString('ITSF')) {
```
Borewit, line 1329: Please move this up to the 4-byte signatures, to prevent unnecessary data to be read.

Borewit (changes requested):

Borewit (approved):
LGTM

PR #448, opened 2021-04-12
Remove moot check for sync word at odd offsets for MPEG detection

### LINKED ISSUE #445: MPEG sync check at odd offset required?
https://github.com/sindresorhus/file-type/blob/fd1e72c8624018fe67a50edcd1557f153260cdca/core.js#L1345

no idea what it's actually supposed to be.

Borewit:
Remove MPEG sync word check add offset 1.

Fix #445.

### DIFF
--- a/core.js
+++ b/core.js
@@ -1341,50 +1341,47 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
-	// Check for MPEG header at different starting offsets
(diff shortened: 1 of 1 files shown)


Borewit:
Also did an integration tests with [music-metadata](https://github.com/Borewit/music-metadata), this change did not break anything.

PR #451, opened 2021-04-15
Add support for VCF (and fix ICS detection)

mdartic:
Link to the wiki page : https://en.wikipedia.org/wiki/VCard

- **One PR per file type.**

### DIFF
--- a/core.js
+++ b/core.js
@@ -911,7 +911,7 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
-	if (checkString('BEGIN:VCALENDAR')) {
+	if (checkString('BEGIN:VCALEN')) { // limit to 12 characters as precised in bytesRead const
 		return {
 			ext: 'ics',
 			mime: 'text/calendar'
--- a/core.js
+++ b/core.js
@@ -918,6 +918,13 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
+	if (checkString('BEGIN:VCARD')) {
+		return {
+			ext: 'vcf',
+			mime: 'text/vcard'
+		};
+	}
+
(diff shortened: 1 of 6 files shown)


Borewit (changes requested):
The file tests are sorted on required initial sample length.
As the sample length now increased you need to move it further down.
That allows progressive reading.

core.js
```diff
@@ -911,13 +911,20 @@ async function _fromTokenizer(tokenizer) {
 		};
 	}
 
-	if (checkString('BEGIN:')) {
+	if (checkString('BEGIN:VCALEN')) { // Limit to 12 characters as precised in bytesRead const
```
Borewit, line 914: Just move it further down, where the sample length is increased.

mdartic: Not sure to really understand, do you want me I just move `BEGIN:VCARD` and `BEGIN:VCALEN` check under the `9-byte signatures` ?
Do you want I add a new `11-byte signature` comment too ?

Borewit: I recommend to move `BEGIN:VCARD` & `BEGIN:VCALENDER` to the appropriate locations.

Alternatively leave it here and keep `BEGIN:`
```js
if (checkString('BEGIN:')) { // So this is 6 byte test, the sub-tests take care of their own reads, so that is not taken into account
  await tokenizer.ignore(6); // advance pointer after 'BEGIN:'
  await tokenizer.readBuffer(buffer, 9, {mayBeLess: true}); // read word after BEGIN:'
  if (checkString('VCARD')) {
    return // vcard type
  } else if (checkString('VCALENDER')) {
    return // vcalander type
  }
  return; // as we permanently read, cannot proceed with other types
}

```

> Do you want I add a new 11-byte signature comment too ?

That is not necessary.

Borewit (approved):

PR #458, opened 2021-05-24
Remove an unnecessary dependency

### LINKED ISSUE #437: Read blob in nodejs
Nodejs have introduced buffer.Blob and we are about to get them from fs soon also

but they have not introduced the FileReader and likely never will. cuz there are some new file reading methods on the blob now with `.stream()`, `.arrayBuffer()`, and `.text()` that can return promises
so i was thinking maybe you could replace this hole function and just use `blob.arrayBuffer()` instead and maybe polyfill `blob.prototype.arrayBuffer()` if it dose not exist

https://github.com/sindresorhus/file-type/blob/2cc0869dcb4381b2223d6759e3f5a8542b1e7975/browser.js#L18-L45

I have also another concern, is it really necessary to read the hole blob into memory?
isn't the first and last 1000 bytes or something only needed?

it would seem more efficient to do something like:
```js
chunk = new Blob([
  blob.slice(0, 4500),
  blob.slice(-4500)
]).arrayBuffer()
```

I would aslo suggest that you use Uint8Array, ArrayBuffer and DataView instead of using nodes Buffer as that would lower the browser bundle and not drag in any of nodes core modules.

I would also suggest that you make your `fromStream` method acceptable of a AsyncIterator that yields Uint8Arrays instead of relying on either node streams or whatwg streams or nodes EventEmitter since it could lower the bundle even a lot further
since node stream are already async iterable. and whatwg streams are too but no browser have implemented it.

jimmywarting:
Would like to remove the FileReader as well, but wasn't sure what browser you support

Closes #437

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,6 +1,5 @@
 'use strict';
 const {ReadableWebToNodeStream} = require('readable-web-to-node-stream');
-const toBuffer = require('typedarray-to-buffer');
 const core = require('./core');
 
 async function fromStream(stream) {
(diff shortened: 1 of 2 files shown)


browser.js
```diff
@@ -11,25 +10,25 @@ async function fromStream(stream) {
 }
 
 async function fromBlob(blob) {
-	const buffer = await convertBlobToBuffer(blob);
+	const buffer = await convertBlobToBuffer(blob).then(Buffer.from);
```
sindresorhus, line 13: Use await instead of `.then`.

Borewit (approved):
Looks good to me

PR #459, opened 2021-06-11
Add support for NTF/NITF format

swis-anno:
support for National Imagery Transmission Format

I could not add unit tests, because of the size of original NTF files, tested locally 

- **One PR per file type.**
- Add the file type detection logic to the `core.js` file.
- Add the file extension to the `Supported file types` section in the readme, in the format ```- [`<extension>`](URL) - Format name```, for example, ```- [`png`](https://en.wikipedia.org/wiki/Portable_Network_Graphics) - Portable Network Graphics```


core.d.ts
```diff
@@ -274,8 +276,9 @@ declare namespace core {
 		| 'application/x-asar'
 		| 'model/stl'
 		| 'application/vnd.ms-htmlhelp'
-		| 'model/3mf'
+		| 'model/3mf'
```
sindresorhus, line 279: Trailing whitespace.

core.d.ts
```diff
@@ -141,6 +141,8 @@ declare namespace core {
 		| '3mf'
 		| 'zst'
 		| 'jxl'
+		| 'ntf'
+		| 'nitf'
```
sindresorhus, line 145: We don't add extensions that aren't returned.

sindresorhus (changes requested):

core.d.ts
swis-anno: trimmed

core.d.ts
swis-anno: removed, 
however, I left a note about 'nitf' in package.json (in keywords section), b/c we have files with both extensions, and for both  above detecting code works fine. Let me know if that is OK with you.

swis-anno:
Hey, please let me know if you want me to change anything more.

core.js
```diff
@@ -1162,6 +1162,13 @@ async function _fromTokenizer(tokenizer) {
 		}
 	}
 
+	if (check([0x4E, 0x49, 0x54, 0x46, 0x30])) {
```
sindresorhus, line 1165: This is in the incorrect place. It should be grouped with the other 5-byte signatures.

sindresorhus:
Tests are failing.

sindresorhus:
@swis-anno Bump :)

PR #464, opened 2021-07-10
Fix `mimeTypes` TypeScript type

charsleysa:


sindresorhus:
You need to update the test too.

charsleysa:
Updated the test.

PR #465, opened 2021-07-10
Update dependency to token-types v3, supporting BigInt

Borewit:
Update dependency to [token-types v3](https://github.com/Borewit/token-types/releases/tag/v3.0.0), supporting [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)

- [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) requires Node version 10, updated module engine API requirement.
- Update [strtok3](https://github.com/Borewit/strtok3) to [v6.1.0](https://github.com/Borewit/strtok3/releases/tag/v6.1.0) to allign the dependent token-types version.

Related: Borewit/token-types#300

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,10 +10,10 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
+          - 16
           - 14
           - 12
           - 10
-          - 8
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


sindresorhus:
Since this has no user-facing changes, I'm not going to do another release until there are other changes.

PR #466, opened 2021-07-20
Update to token-types 4.0.1 & strtok3 6.2.2

Borewit:
Changes:
1. Update to [token-types 4.0.1](https://github.com/Borewit/token-types/releases/tag/v4.1.0) & [strtok3 6.2.2](https://github.com/Borewit/strtok3/releases/tag/v6.2.2)
1. The updated dependencies try to abstract Buffer dependency by using `Uint8Array` where possible.
1. Remove `Buffer` as type, as `Buffer` is a sublcass of `Uint8Array`.
1. Add [@tokenizer/token](https://github.com/Borewit/tokenizer-token) to development dependencies. This was not required in the past as [@tokenizer/token](https://github.com/Borewit/tokenizer-token) was incorrectly included indirectly as a normal dependency.

Dependent PR: Borewit/music-metadata#852, to align dependencies [token-types](https://github.com/Borewit/token-types) & [strtok3](https://github.com/Borewit/strtok3)

Dependency on `Buffer` is still there, but in some dependent module the `Buffer` dependency has been entirely abstracted to `Uint8Array`. Let's say I try to get rid of Buffer dependencies if not really required.

By doing this, I hope on the long run I hope modules will be better suitable for multipurpose (Node.js & browser) usage.

Fixes:
* #469

### DIFF
--- a/core.js
+++ b/core.js
@@ -20,11 +20,11 @@ async function fromStream(stream) {
 }
 
 async function fromBuffer(input) {
-	if (!(input instanceof Uint8Array || input instanceof ArrayBuffer || Buffer.isBuffer(input))) {
+	if (!(input instanceof Uint8Array || input instanceof ArrayBuffer)) {
 		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\` or \`ArrayBuffer\`, got \`${typeof input}\``);
 	}
 
-	const buffer = input instanceof Buffer ? input : Buffer.from(input);
+	const buffer = input instanceof Uint8Array ? input : new Uint8Array(input);
 
 	if (!(buffer && buffer.length > 1)) {
 		return;
--- a/core.d.ts
+++ b/core.d.ts
@@ -301,10 +301,10 @@ declare namespace core {
 
 	If file access is available, it is recommended to use `.fromFile()` instead.
 
-	@param buffer - A buffer representing file data. It works best if the buffer contains the entire file, it may work with a smaller portion as well.
+	@param buffer - An Uint8Array or Buffer representing file data. It works best if the buffer contains the entire file, it may work with a smaller portion as well.
 	@returns The detected file type and MIME type, or `undefined` when there is no match.
 	*/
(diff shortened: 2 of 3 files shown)


Borewit:
@jimmywarting, can you please review?

core.js
```diff
@@ -20,11 +20,11 @@ async function fromStream(stream) {
 }
 
 async function fromBuffer(input) {
-	if (!(input instanceof Uint8Array || input instanceof ArrayBuffer || Buffer.isBuffer(input))) {
+	if (!(input instanceof Uint8Array || input instanceof ArrayBuffer)) {
 		throw new TypeError(`Expected the \`input\` argument to be of type \`Uint8Array\` or \`Buffer\` or \`ArrayBuffer\`, got \`${typeof input}\``);
 	}
 
-	const buffer = input instanceof Buffer ? input : Buffer.from(input);
+	const buffer = input instanceof Uint8Array ? input : new Uint8Array(input);
```
jimmywarting, line 27: Had to take a extra look at this, but then i realized that you guard yourself with type checking above.

`Buffer.from(input)` dose a little bit more magic than `new Uint8array(input)` like converting strings.
if this is something you also want to support, then a easy way to convert string to uint8 is via `new TextEncoder().encode(str)` 
I just released a package that converts literally anything to Uint8Array with 0 copies if it's of any interest to you... it's called [to-uint8array](https://www.npmjs.com/package/to-uint8array)


----

The `Buffer.isBuffer(input)` would never be called cuz it would be covered by instanceof uint8 check

jimmywarting (approved):

core.js
sindresorhus: > like converting strings.

We wouldn't want that at least. The API is documented to accept a buffer.

Borewit: So we allow kind of 3 different input's: Buffer, Uint8Array, ArrayBuffer.
I don't think we need to explicitly add a Buffer here,  as Uint8Array is sufficient.

As @sindresorhus indicated, `string` was not something we had in mind as an input.

jimmywarting: I didn't expect it to work with strings either.

Just wanted to recommend my package cuz it can convert any ArrayBufferViews (including Buffer + DataView) & ArrayBuffer and turn them into uint8arrays 
Quite handy for any typed array

Issue #467, opened 2021-07-20
`stream()` may throw End-Of-Stream error


Borewit:
Method `stream()` may throw End-Of-Stream error.

Working on #434, in which I make the sample-size configurable this became very easy reproachable by making the sample size smaller (e.g. 30 bytes) and running it against the fixture.

What the behavior should be, is that the file-type cannot be determined. The End-Of-Stream is an error which is kind of likely to happen as we the underlying mechanisms are running out out of data due to the sample length limitation.

Borewit:
Fixed by #468

PR #468, opened 2021-07-20
Prevent End-Of-Stream error in `stream()`

Borewit:
Suppress the End-Of-Stream, instead assign `undefined` to `stream.fileType`.
Restrict updating strtok3 to minor version.

Fixes: 
* #467
* #469

### DIFF
--- a/core.js
+++ b/core.js
@@ -1436,7 +1436,6 @@ const stream = readableStream => new Promise((resolve, reject) => {
 			const fileType = await fromBuffer(chunk);
 			pass.fileType = fileType;
 		} catch (error) {
-			reject(error);
 			if (error instanceof strtok3.EndOfStreamError) {
 				pass.fileType = undefined;
 			} else {
--- a/core.js
+++ b/core.js
@@ -1437,6 +1437,11 @@ const stream = readableStream => new Promise((resolve, reject) => {
 			pass.fileType = fileType;
 		} catch (error) {
 			reject(error);
+			if (error instanceof strtok3.EndOfStreamError) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
@Borewit Are you ok with us moving `file-type` to ESM? 

That would require Node.js 12, but would further improve browser usage.

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

Borewit:
> @Borewit Are you ok with us moving `file-type` to ESM?
> 
> That would require Node.js 12, but would further improve browser usage.
> 
> https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

I need to get my head around the ESM modules. As you know I have been working on this Buffer abstraction and `UInt8Array`, requiring Node v10. I would like to fully finalize that effort, and shake issue if required, before refactoring any further.
Involves:
* [strok3](https://github.com/Borewit/strtok3)
* [token-types](https://github.com/Borewit/token-types)
* [peek-readable](https://github.com/Borewit/peek-readable)
* [file-type](https://github.com/sindresorhus/file-type)
* [@tokenizer/s3](https://github.com/Borewit/tokenizer-s3)
* [tokenizer-range](https://github.com/Borewit/tokenizer-range)
* [music-metadata](https://github.com/Borewit/music-metadata)
* [music-metadata-browser](https://github.com/Borewit/music-metadata-browser)

But it sounds good to me, I looking forward to the day these modules are equally easy to use in Node.js and browser environment.
Life is too short to deal with over-complicated things.

PR #476, opened 2021-07-29
Fix file info size check

LinusU:
Fixing the failing tests as requested here https://github.com/sindresorhus/file-type/pull/472#discussion_r679383277 😎

### DIFF
--- a/core.js
+++ b/core.js
@@ -74,7 +74,7 @@ async function _fromTokenizer(tokenizer) {
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
 	// Keep reading until EOF if the file size is unknown.
-	if (tokenizer.fileInfo.size === undefined || tokenizer.fileInfo.size === null || tokenizer.fileInfo.size === 0) {
+	if (tokenizer.fileInfo.size === undefined) {
 		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;
 	}
 
--- a/core.js
+++ b/core.js
@@ -74,7 +74,7 @@ async function _fromTokenizer(tokenizer) {
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
 	// Keep reading until EOF if the file size is unknown.
-	if (tokenizer.fileInfo.size === 0) {
+	if (tokenizer.fileInfo.size === undefined || tokenizer.fileInfo.size === null || tokenizer.fileInfo.size === 0) {
 		tokenizer.fileInfo.size = Number.MAX_SAFE_INTEGER;
 	}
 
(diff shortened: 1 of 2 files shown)


core.js
```diff
@@ -74,7 +74,7 @@ async function _fromTokenizer(tokenizer) {
 	const checkString = (header, options) => check(stringToBytes(header), options);
 
 	// Keep reading until EOF if the file size is unknown.
-	if (tokenizer.fileInfo.size === 0) {
+	if (tokenizer.fileInfo.size === undefined || tokenizer.fileInfo.size === null || tokenizer.fileInfo.size === 0) {
```
LinusU, line 77: It's specifically `undefined` and not `null`, but I figured it would be better to accept both?

It's also never `0` (in the tests at least), I think before they are using `undefined` to indicate that the size hasn't been set yet. So potentially we want to remove `0`? 🤔

sindresorhus: @Borewit The `.size` property is not documented to be `undefined`. The docs should use `number?` instead of `number` to make that clear.

sindresorhus: @LinusU Let's only check for `undefined`.

sindresorhus:
Thank for looking into this 🙌

core.js
Borewit: I just started looking into the failing tests as well. I noticed 2 things:

1. The `fileInfo.size` is not for some reason not set reading from a file-stream (`fs.createReadStream`) includes the original file path which is used to determine the file size)
2. There are cases `fileInfo.size` is `undefined` (because the stream length is not known), so that code needs to be improved.

I don't directly remember where the `0` is coming from. I will try to give some clarity on that and improve the documentation accordingly.

Borewit: Issue with size being 0 introduced here: https://github.com/sindresorhus/file-type/blame/331b1620c0d6a29b70a2e8e6c246cb2bdd93b183/core.js#L73

sindresorhus: > Issue with size being 0 introduced here: https://github.com/sindresorhus/file-type/blame/331b1620c0d6a29b70a2e8e6c246cb2bdd93b183/core.js#L73

Yes, the linter auto-fix assumed `.size` would be a number and never `undefined`. It doesn't read the types. The previous check was too loose anyway though, so it's good that we make it more strict now.

Borewit:
I am puzzled, so we went from:

```js
if (!tokenizer.fileInfo.size) // works, but linter does not like it
```
to
```js
if (!tokenizer.fileInfo.size === 0) // does not work
```
to
```js
if (tokenizer.fileInfo.size === undefined || tokenizer.fileInfo.size === null || tokenizer.fileInfo.size === 0) // could be simplified to...
```
Are we working for the linter or the linter for us?

LinusU:
@Borewit I actually think that the last one of those is the better since when reading the code it's obvious that the value could be `null`, `undefined`, or `0`. The original code didn't express the intent clear at all, which is what led to the problem in the first place. I think it's better to optimise code for reading and for clarity, rather than brevity...

PR #482, opened 2021-08-13
Improve parsing TIFF files

### LINKED ISSUE #480: DNG is falsely detected as TIFF
I have a couple of test images I've been working with to test something I'm working on, which uses the file-type library. All of my DNG images seem to get falsely picked up as TIFF images (even though it looks like there's an explicit check for DNG in the code).

I've tested with images from three different cameras (a Leica M10, a Nikon D5100, and a Pentax K-70 to be specific.) and would be happy to provide some files.

Borewit:
Changes:
- Smarter TIFF detection smarter, by actually reading the TIFF tags and their contents.
- Improve Big-Endian encoded TIFF detection. 
- Replaced some TIFF fixtures with larger samples 
- Basic Big TIFF file detection (same extension), no support for specialized type yet

The existing samples are truncated, so these are kind of stuck with old school magic byte pattern matching. 
Therefor I replaced them with original files.

The updated Sony file type `.arw` has been a bit of guess. It basically looks for a tag, which looks like a Sony private tag. Other `.arw` may or may not have this tags.

~~**Warning**: fixtures are very large, 25 MB each!!!~~
Truncated samples to 100.000 bytes each. This is enough to read it's tags.

Fixes #480

Source sample file `.arw`:  https://www.dpreview.com/sample-galleries/9547785333/


sindresorhus:
Nice work! 👍

Issue #485, opened 2021-08-30
Type detection failing for webm format file


Abhinav-grd:
I have a ubuntu screen Recording File of webm format for which type detection fails

File - https://drive.google.com/file/d/1g14wtH7qNpzfA1eMVqYCjPHBgQhs8nNU/view?usp=sharing

I am using ```FileType.fromBuffer(filedata);```

Borewit:
Doc-type is incorrectly read due to trailing null-octets, which should be supported according the standard.

Ref: [EBML 11.1.3. Terminating Element Data](https://tools.ietf.org/id/draft-ietf-cellar-ebml-03.html#terminating-element-data)

PR #486, opened 2021-08-30
Improve WebM detection

Borewit:
Fixes recognition of WebM format.

Fixes: sindresorhus/file-type#485


test.js
```diff
@@ -203,6 +203,9 @@ const names = {
 		'fixture-fast-web', // PDF saved from Adobe Illustrator, using the default "[Illustrator Default"] preset, but enabling "Optimize for Fast Web View"
 		'fixture-printed', // PDF printed from Adobe Illustrator, but with a PDF printer.
 	],
+	webm: [
+		'fixture-null', // EBML DocType with leading null character
```
sindresorhus, line 207: The issue and the code says "trailing", but the pull request title and comment here says "leading".

core.js
```diff
@@ -736,7 +736,8 @@ async function _fromTokenizer(tokenizer) {
 			while (children > 0) {
 				const element = await readElement();
 				if (element.id === 0x42_82) {
-					return tokenizer.readToken(new Token.StringType(element.len, 'utf-8')); // Return DocType
+					const rawValue = await tokenizer.readToken(new Token.StringType(element.len, 'utf-8'));
+					return rawValue.replace(/\00.*$/g, ''); // Return DocType
```
sindresorhus, line 740: Does it document the maximum amount of null characters there could be? Would be nice to have a limit in place so it wouldn't hang on faulty files that has too many null characters.

Borewit: There is no maximum, it's used as a kind of padding. 
The maximum length is of the string read is already terminated by `element.len`.

Borewit: Ah you mean check `element.len`? It can be 2^35 (exceeding `number`) and the length is encoded a specific way ([VINT examples](https://tools.ietf.org/id/draft-ietf-cellar-ebml-03.html#vint-examples)). 
At that point the assumption is already it is EBML and starts to consume the tokenizer
and iterate through the [_EBML elements_](https://tools.ietf.org/id/draft-ietf-cellar-ebml-03.html#notation-and-conventions).
You could say, the docType must be a relative short value, but then we already matched `0x1A, 0x45, 0xDF, 0xA3` & `0x42_82`. Extremely unlikely we hit that point without the format being EBML.

PR #493, opened 2021-09-19
Refactor function `_fromTokenizer` to a `FileTypeParser` class

Borewit:
Prevent nesting of functions and utilize class declaration (since ES6).
Class `FileTypeParser` hold stateful objects and variables throughout the parsing process,
and a list of private function depending on that.

### DIFF
--- a/core.js
+++ b/core.js
@@ -55,7 +55,7 @@ function _check(buffer, headers, options) {
 
 export async function fileTypeFromTokenizer(tokenizer) {
 	try {
-		return _fromTokenizer(tokenizer);
+		return new FileTypeParser().parse(tokenizer);
 	} catch (error) {
 		if (!(error instanceof strtok3.EndOfStreamError)) {
 			throw error;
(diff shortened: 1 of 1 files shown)


sindresorhus:
This looks like a classic misuse of classes. There's really no state. It's only the input buffer which is static.

Borewit:
> This looks like a classic misuse of classes. There's really no state. It's only the input buffer which is static.

`buffer` is not static, it's assigned to a function scope. And so is `bytesRead`. As multiple function require access to these, it makes perfect sense to put them in a class structure. As a class is not much more then a special function designed for holding functions and declarations it makes perfect sense, because that is exactly what `_fromTokenizer` does, it holds functions and declarations.

sindresorhus:
You're right about `buffer`, but `bytesRead` is only used once right below its declaration. There's no reason it should be exposed and mutable to the whole class. This is the kind of problems you get with a class. It's too easy to add everything to `this` and suddenly everything is mutable from anywhere. With functions, you have to explicitly pass in what you want to expose/mutate. I'm not saying we shouldn't use classes, but it's something to consider.

I just fail to see how it simplifies things in the current PR. It lets us not pass `buffer` explicitly to a couple of functions. That's it. I would probably see more benefit to this change if you took advantage and split up `parse()` into smaller private methods. For example, a method for each byte signature length, so `parse()` wouldn't be so huge.

Borewit:
> It's too easy to add everything to this and suddenly everything is mutable from anywhere. 

Same for the function scope, and I find that harder to debug. Maybe me.

> With functions, you have to explicitly pass in what you want to expose/mutate. 

I see what you mean. Yet if we pass the context as a nested first argument, it is not much different then a class. That is how we kind of emulated objects in `C`.

> I just fail to see how it simplifies things in the current PR.

I agree it is not a drastic improvement. I think it slight worse then the previous structure.   
 
> . I would probably see more benefit to this change if you took advantage and split up parse() into smaller private methods. For example, a method for each byte signature length, so parse() wouldn't be so huge.

Yeah I agree, I see some challenges making it so, but it is far from ideal the way it is now.

sindresorhus:
> I see some challenges making it so

One idea would be to abuse errors:

```js
class FileTypeParser {
	check(header, options) {
		return _check(this.buffer, header, options);
	}

	checkString(header, options) {
		return this.check(stringToBytes(header), options);
	}

	async parse(tokenizer) {
		// ...

		try {
			_parse2Byte();
			_parse3Byte();
		} catch (error) {
			return error.value;
		}
	}

	_parse2Byte() {
		if (this.check([0x42, 0x4D])) {
			throw Result({
				ext: 'bmp',
				mime: 'image/bmp',
			});
		}

		if (this.check([0x0B, 0x77])) {
			throw Result({
				ext: 'ac3',
				mime: 'audio/vnd.dolby.dd-raw',
			});
		}
	}
}
```

That way we don't have to check the value after each call.

Alternatively, we could use a generator or async generator: https://gist.github.com/mafintosh/9cfb78ec8f2e4989ecd76d1b005d866e

Borewit:
I prefer to keep the throw for actual errors.

We can reasonable effective check the return type in a loop: 
```js
const fileType = [
  this._parse2Byte, 
  this._parse3Byte
].find(parse => parse());
```

or

```js
const fileType = [2, 3].find(parse => this['parse' + bufLen + 'Byte']());
```

At the other end, the way it is now, it is kind of dictionary. There is not to much code at each file type test, each entry is reasonable to the point.

sindresorhus:
> We can reasonable effective check the return type in a loop:

From experience, the problem with that is when you need to have some adhoc logic between the steps for example. But I guess we can deal with that if it happens.

core.js
```diff
@@ -55,1407 +55,1421 @@ function _check(buffer, headers, options) {
...
 
-	await tokenizer.peekBuffer(buffer, {length: bytesRead, mayBeLess: true});
+	checkString(header, options) {
+		return this.check(stringToBytes(header), options);
+	}
 
-	// -- 2-byte signatures --
+	async parse(tokenizer) {
+		this.buffer = Buffer.alloc(minimumBytes);
+		this.bytesRead = 12;
```
sindresorhus, line 77: Can you change this to not be on `this.` since it's only used in this scope?

Borewit: It's not only used in that scope, it's also by `this.check()`:

https://github.com/sindresorhus/file-type/blob/75a69b4d64a6f277c8ec628aed814caf33c2de2a/core.js#L67-L69

sindresorhus: This comment is attached to `this.bytesRead = 12;` and is what I am talking about. Not `this.buffer`.

PR #497, opened 2021-10-04
Detect both raw and BDAV versions of MPEG-2 Transport Streams

pasieronen:
MPEG-2 Transport Stream specification (Section 2.3 and Annex T):
https://www.itu.int/rec/dologin_pub.asp?lang=f&id=T-REC-H.222.0-201703-S!!PDF-E&type=items

Blu-ray Disc Audio-Video (BDAV) MPEG-2 Transport Stream (Section 3.1):
http://www.blu-raydisc.com/Assets/downloadablefile/2b_bdrom_audiovisualapplication_0305-12955-15269.pdf
 
The "raw" fixture file was generated from the existing BDAV fixture with this command (`-mpegts_m2ts_mode 0` is the relevant part):

    ffmpeg -i fixture-bdav.mts -f mpegts -vcodec copy -acodec copy -mpegts_m2ts_mode 0 -t 1 fixture-raw.mts


Borewit (approved):
Looks good to me

PR #507, opened 2021-12-14
Update strtok3 & token-types for explicit `node:buffer` imports

### LINKED ISSUE #502: node: prefix error using in browser
Get the following errors when I try to use fileTypeFromBlob in browser:
```
ERROR in node:buffer
Module build failed: UnhandledSchemeError: Reading from "node:buffer" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
...
```
```
ERROR in node:stream
Module build failed: UnhandledSchemeError: Reading from "node:stream" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
...
```
I can successfully build my project only with `target: "node"` in webpack config. Webpack 5.64.4, file-type 17.0.0.

Borewit:
Update "strtok3" to ^[7.0.0-alpha.5](https://github.com/Borewit/strtok3/releases/tag/v7.0.0-alpha.5)
Update "token-types" to ^[5.0.0-alpha.0](https://github.com/Borewit/token-types/releases/tag/v5.0.0-alpha.0) (first ES module).

Fixes #502

### DIFF
--- a/core.js
+++ b/core.js
@@ -1,5 +1,5 @@
 import {Buffer} from 'node:buffer';
-import Token from 'token-types';
+import * as Token from 'token-types';
 import * as strtok3 from 'strtok3/core';
 import {
 	stringToBytes,
--- a/package.json
+++ b/package.json
@@ -196,8 +196,8 @@
 	],
 	"dependencies": {
 		"readable-web-to-node-stream": "^3.0.2",
-		"strtok3": "7.0.0-alpha.4",
-		"token-types": "^4.1.1"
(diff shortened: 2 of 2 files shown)


sindresorhus:
Seems to be failing on Node.js 12

Borewit:
> Seems to be failing on Node.js 12

Yeah, but that is not specific in this PR. I can reproduce the error, but not outside av.

PR #509, opened 2021-12-15
Supported files list in readme.md sorted

ZE0TRON:
Hi, the list of the supported file in the readme.md is not sorted. Today I have discovered this library and was trying to figure out if it supports pdfs and some other format it was hard to find them in the list. It's a small change but I think it could help the people in my position. If there is a particular reason for the current order you may just ignore this pr.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -318,144 +318,144 @@ Returns a `Set<string>` of supported MIME types.
 
 ## Supported file types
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I'm open to it, but it should be automatically enforced first: https://github.com/sindresorhus/file-type/issues/511

PR #512, opened 2021-12-17
Add support for AVIF Sequence

tk-wing:
This PR adds support for the`image/avif-sequence` such as animation.

`avif-sequence` mime-type has removed, as described [here](https://github.com/AOMediaCodec/av1-avif/pull/86).
So, `image/avif-sequence` should be used `image/avif` mime-type.

[FileTypeBox Constraint forAVIF](https://aomediacodec.github.io/av1-avif/#file-constraints)


core.js
```diff
@@ -477,6 +477,7 @@ class FileTypeParser {
 			const brandMajor = this.buffer.toString('binary', 8, 12).replace('\0', ' ').trim();
 			switch (brandMajor) {
 				case 'avif':
+				case 'avis':
```
Borewit, line 480: Are you sure this is right extension and MIME-type for AVIF Sequence file?
What about `.avifs` and `image/avif-sequence`

Ref:
- http://fileformats.archiveteam.org/wiki/AVIF
- https://github.com/AOMediaCodec/libavif/issues/112)?

Borewit (changes requested):

core.js
Borewit: Never mind, found the answer: https://github.com/AOMediaCodec/av1-avif/issues/59#issuecomment-645401286

Borewit (approved):

PR #513, opened 2021-12-20
Fix error: Cannot find module 'strtok3/core'

acardinale:
Set the correct path of the core file of the strtok3 package
https://github.com/Borewit/strtok3/blob/v7.0.0-alpha.7/lib/core.ts


acardinale:
i wrong

irux:
why did you close this ? @acardinale

kryopix:
hi, this error is still in latest release, when will be the next release with this issue fixed, so i can update my dependencies?

PR #516, opened 2021-12-27
Add support for COFF (Unix Common Object File Format)

votruongan:
**File magic number reference:** [https://en.wikipedia.org/wiki/COFF](https://en.wikipedia.org/wiki/COFF)

**MIME type:** I have looked up in [IANA's list](https://www.iana.org/assignments/media-types/media-types.xhtml) and [tika supported MIME types](https://tika.learnworlds.com/mime-types), but couldn't find any MIME type associated with COFF, so I left `application/octet-stream` as default.

- **One PR per file type.**


sindresorhus:
Do you actually need this? Wikipedia says the format is mostly replaced by ELF.

votruongan:
No, my needs haven't touched any COFF file yet. I am just adding support for as much executable files as possible.

COFF is deprecated, but its extended version, like XCOFF (different magic number) is still able to run on IBM AIX systems, so I think I would also open a PR for XCOFF.

So, supporting COFF in this package is much more about diversity and historical purpose. other types of COFF would have more practical impacts.

sindresorhus:
> So, supporting COFF in this package is much more about diversity and historical purpose.

The goal of this package is to support commonly used formats, not everything. For historical or little used formats, it's better to just spawn the `file` command.

sindresorhus:
> COFF is deprecated, but its extended version, like XCOFF (different magic number) is still able to run on IBM AIX systems, so I think I would also open a PR for XCOFF.

I'm going to pass on that one too. It's enough work to maintain popular modern formats. I don't want to open the flood gates for everything out there.

Issue #440, opened 2021-03-10
Support DWG File


bireycikan:
[DWG file](https://en.wikipedia.org/wiki/.dwg)

Filename extension: dwg, dws, dwt
Media type: application/acad

Magic numbers: 41 43

Ephigenia:
@bireycikan I’ve opened a PR for this file format and also tested it on a random DWG file from the net which works fine.

While verifying that 41 43 is correct I also found this list here: https://gist.github.com/ldong/f334b9dcd421c99e094d which states that 41 43 31 30 is also correct.

somebody1234:
re: media type... wikipedia says `image/vnd.dwg`, [autodesk](https://forge.autodesk.com/ja/node/1650) says `https://forge.autodesk.com/ja/node/1650`

somebody1234:
as for the headers for DWGs - it seems to be just the version numbers on the [wikipedia page for dwg](https://en.wikipedia.org/wiki/.dwg) - in the case of AC10XX versions (well, AC1015 and above - not sure where you'd be able to find files older than that), anyway

somebody1234:
relevant:
[libredwg tests for header](https://github.com/LibreDWG/libredwg/blob/280ef56d1b38298947e1cd5d43edfb68203845e2/test/unit-testing/common_test.c#L79-L92)
[raw headers in libredwg](https://github.com/LibreDWG/libredwg/blob/f46b3f0fa33bc4359a8a3f0c120ec54d84d6f8b7/src/common.c#L33-L91)

PR #520, opened 2021-12-29
Add support for DWG

Ephigenia:
This PR adds support for [DWG](https://en.wikipedia.org/wiki/.dwg) files according to the issue https://github.com/sindresorhus/file-type/issues/440 and the [File Signature](https://www.garykessler.net/library/file_sigs.html)

Filename extension: `dwg`, `dws`, `dwt`
Media type: `application/acad`
magic numbers: `41 43`


core.js
```diff
@@ -994,6 +994,13 @@ class FileTypeParser {
 			};
 		}
 
+		if (this.check([0x41, 0x43])) {
```
Borewit, line 997: You mention `41 43 31 30` could work as well. 
Why don't you use that one? It would decreases the risk on a false positive significant.

Borewit (commented):
Please try to link to references.
Use official source, standards, wiki to verify magic numbers, mime-type found in other code.

core.js
Ephigenia: sure, changed the line to use the longer sequence `41 43 31 30`

Ephigenia:
Mmh, according to the [File Signatures](https://www.garykessler.net/library/file_sigs.html) it looks safer to go for the shorter 2 number-version (`41 43`) as the following numbers seem to indicate the version of the file. I changed it back to the 2 number version.

core.js
Borewit, line 997: The [file signatures](https://www.garykessler.net/library/file_sigs.html) also shows that `41 43` is used by other file types (ACL & ACSD), so just testing on `41 43` and assume it's an AutoCAD file is a tricky assumption.

I propose to test on the following: `41 43 31 2E`, `41 43 32 2E`, `41 43 31 30`, where you can keep the initial test on `41 43`. After that you test on `31 2E`, `32 2E` & `31 30` at `offset=2`.

Please move the test to ~line 883, as the last test of the 4 byte signature group.

Borewit (changes requested):
The [file signatures](https://www.garykessler.net/library/file_sigs.html) also shows that `41 43` is used by other file types (ACL & ACSD), so testing on `41 43` is not very reliable.

I propose to test on the following: `41 43 31 2E`, `41 43 32 2E`, `41 43 31 30`, where you can keep the initial test on `41 43`.

core.js
Borewit: The [DWG specification](https://www.opendesign.com/files/guestdownloads/OpenDesign_Specification_for_.dwg_files.pdf) explains the header in 3.2. The header is ASCII, and you can make use of that.

The `41 43` is actually `AC` (**A**uto**C**AD I guess). 
The 4 bytes after that give you the version. You can test of that if version is in [1000, 1050]. (The standard describes a range from [1012, 1032]. So there is some backward and future compatibility here.

That makes it a 6 bytes check.

Borewit:
Do you mind finalizing your PR @Ephigenia?

Ephigenia:
@Borewit yes, I will in the next few days, sorry for the delay!

sindresorhus:
@Ephigenia Friendly bump :)

Issue #527, opened 2022-01-19
PDF is not detected correctly if it starts with a UTF-8 BOM byte sequence


tamas-eppel-mediamarkt:
Hello,

I have a PDF generated by `Oracle11gR1 AS Reports Services` - `Oracle PDF driver`. We receive this from a third party. This document has three additional bytes at the beginning of the file, and because of this the file-type check fails: returns undefined. However the linux `file` command, or a PDF viewer recognizes the file.

I have attached a file, which shows this problem.

```
❯ file ~/Downloads/test.pdf
/home/teppel/Downloads/test.pdf: PDF document (UTF-8), version F.1, 1 pages (zip deflate encoded)
```

[test.pdf](https://github.com/sindresorhus/file-type/files/7898147/test.pdf)

Borewit:
The PDF file starts with the [UTF-8 BOM byte sequence](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8): `0xEF,0xBB,0xBF`.

issuefiler:
A 8-MB PDF file I’ve got starts with

```
0A255044462D312E340D0A25E2E3CFD30D0A312030206F626A0D0A3C3C0D0A2F54797065202F5061
67650D0A2F4D65646961426F78205B20302030203539352E3332203834312E3932205D0D0A2F5265
736F7572636573203C3C202F466F6E74203C3C202F4631203320302052202F463220342030205220
2F4633203520302052202F4634203620302052203E3E202F4578744753746174650D0A3C3C202F47
5337203720302052202F475338203820302052203E3E202F584F626A656374203C3C202F496D6167
6539203920302052202F496D6167653131203130203020520D0A2F496D6167653133203131203020
52202F496D616765313520313220302052202F496D616765313720313320302052202F496D616765
313920313420302052202F496D61676532310D0A313520302052202F496D61676532332031362030
```
```

%PDF-1.4
%����
1 0 obj
<<
/Type /Page
/MediaBox [ 0 0 595.32 841.92 ]
/Resources << /Font << /F1 3 0 R /F2 4 0 R /F3 5 0 R /F4 6 0 R >> /ExtGState
<< /GS7 7 0 R /GS8 8 0 R >> /XObject << /Image9 9 0 R /Image11 10 0 R
/Image13 11 0 R /Image15 12 0 R /Image17 13 0 R /Image19 14 0 R /Image21
15 0 R /Image23 16 0
```

[...]

PR #529, opened 2022-01-30
UTF8 BOM recursion

Borewit:
Resolves sindresorhus/file-type#527

- Strip of UTF-8 BOM with a recursion  
- Add support field text files preceded with [Byte-Order-Marking](https://en.wikipedia.org/wiki/Byte_order_mark).

### DIFF
--- a/core.js
+++ b/core.js
@@ -151,6 +151,12 @@ class FileTypeParser {
 
 		// -- 3-byte signatures --
 
+		if (this.check([0xEF, 0xBB, 0xBF])) { // UTF-8-BOM
+			// Strip off UTF-8-BOM
+			this.tokenizer.ignore(3);
+			return this.parse(tokenizer);
+		}
+
 		if (this.check([0x47, 0x49, 0x46])) {
 			return {
 				ext: 'gif',
@@ -1021,13 +1027,6 @@ class FileTypeParser {
 			};
 		}
 
-		if (this.check([0xEF, 0xBB, 0xBF]) && this.checkString('<?xml', {offset: 3})) { // UTF-8-BOM
-			return {
-				ext: 'xml',
-				mime: 'application/xml',
-			};
-		}
-
 		// -- 9-byte signatures --
 
 		if (this.check([0x49, 0x49, 0x52, 0x4F, 0x08, 0x00, 0x00, 0x00, 0x18])) {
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't think we should support detecting UTF8 BOM text files. BOM in UTF8 is an anti-pattern:

> Use of a BOM is neither required nor recommended for UTF-8, but may be encountered in contexts where UTF-8 data is converted from other encoding forms that use a BOM or where the BOM is used as a UTF-8 signature. 

And it's a slippery slope for people to keep requesting full support for detecting plain text, not just with BOM.

Borewit:
> BOM in UTF8 is an anti-pattern:

Using a BOM field is not an anti-pattern, it exist for a good reason, it is a standardized magic signature to detect the UTF-16, UTF-8, or UTF-32 & byte order of Unicode encoding. What they mean with _"Use of a BOM is neither required nor recommended for UTF-8"_ is that they don't have an opinion it should or should not be used. If the text encoding may be variable, and not specified otherwise, it is recommended to use it. That is where is designed for. [This misunderstanding is clarified here](https://www.unicode.org/L2/L2021/21038-bom-guidance.pdf).

But if it is ant-pattern or not is completely irrelevant as we are not designing a new file format, we just detect what based on existing specification, with or without BOM field.

> And it's a slippery slope for people to keep requesting full support for detecting plain text, not just with BOM.

I know. Not my aim to change that, but this is what we can derive reasonably reliable from magic binary byte signatures.

Borewit:
But I have seen the BOM field popping up at a lot of places where it definitely should not be, for example in ID3v2 tag values, and that is that is indeed a common made mistake. You indeed cannot just inject a BOM field because it is text.

core.js
```diff
@@ -1165,13 +1168,17 @@ class FileTypeParser {
 			};
 		}
 
-		if (
-			this.check([0xFE, 0xFF, 0, 60, 0, 63, 0, 120, 0, 109, 0, 108]) // UTF-16-BOM-LE
-			|| this.check([0xFF, 0xFE, 60, 0, 63, 0, 120, 0, 109, 0, 108, 0]) // UTF-16-BOM-LE
-		) {
+		if (this.check([0xFE, 0xFF])) { // UTF-16-BOM-LE
```
sindresorhus, line 1171: This is used in multiple places, so instead of a comment, I would make a `UTF16_BOM_LE` constant.

sindresorhus:
I still think supporting `text/plain` is a mistake and it's not even useful as 99.999% of cases do not use BOM. It's just going to cause a lot of support headache.

sindresorhus:
Bump :)

Borewit:
I still think there is no good reason to not support BOM prefix detection. It is already part of other signature detection's, why not for this one? BOM complies with a binary magic byte signature, used to indicate the text encoding. 

That we do not support text base format is in reality not completely true, there are already many text based formats file-type has implemented, in general those who have some kind of delectable signature. PDF is one of those. Some of the PDF sub-types are are on slippery, with very fuzzy recognition patterns rather then steady signatures.

sindresorhus:
> I still think there is no good reason to not support BOM prefix detection. It is already part of other signature detection's, why not for this one? BOM complies with a binary magic byte signature, used to indicate the text encoding.

Just because we can does not mean we should. I'm not against the BOM prefix detection. I think it should be kept, just return `undefined`, so to reduce false-positives. I'm against exposing `text/plain`, as I have already argued. I have a lot of experience with open source and I know declaring support for some plain text is going to create a lot of support overhead. It will not be worth it.

> That we do not support text base format is in reality not completely true, there are already many text based formats file-type has implemented

I don't see where I said we don't support text-based formats. I'm talking about plain text.

Borewit:
Removed all plain text traces .

sindresorhus:
The fixtures either need to be added to test.js somehow or be removed.

PR #536, opened 2022-03-15
Add support for zlib

dr-peppe:
Adding support for zlib compressed files. Currently I've added the magic numbers using "no preset dictionary" as they seem to be the ones used in practice.

I used these as references for the magic numbers:
https://en.wikipedia.org/wiki/List_of_file_signatures
https://stackoverflow.com/questions/9050260/what-does-a-zlib-header-look-like
https://datatracker.ietf.org/doc/html/rfc6713
https://datatracker.ietf.org/doc/html/rfc1950

I did notice a conflict with the magic bytes used for the dmg format. After some investigation I think it's using incorrect detection. E.g. when looking at https://en.wikipedia.org/wiki/List_of_file_signatures it's clear that the format signature is at the end of such files (end–512) which seems to match some of my local dmg files as well. I think this should be resolved in a separate issue.


core.js
```diff
@@ -144,6 +144,18 @@ class FileTypeParser {
 			};
 		}
 
+		if (
+			this.check([0x78, 0x01])
+			|| this.check([0x78, 0x5E])
+			|| this.check([0x78, 0x9C])
+			|| this.check([0x78, 0xDA])
+		) {
```
Borewit, line 152: The magic signatures represent:
```
 CMF |  FLG
0x78 | 0x01 - No compression/low
0x78 | 0x5E - Fastest compression
0x78 | 0x9C - Default compression
0x78 | 0xDA - Best compression 
```
The assumptions here is:
- CM (Compression method) = 8 = Deflate (this assumption seems solid)
- CINFO (Compression info) = 7 = 32K window size
- FDICT (preset dictionary) = 0 = not present

I wonder if we could cover corner cases utilizing:
> The FCHECK value must be such that CMF and FLG, when viewed as
>          a 16-bit unsigned integer stored in MSB order (CMF*256 + FLG),
>          is a multiple of 31.

Ref: https://www.rfc-editor.org/rfc/rfc1950#page-5

Borewit (changes requested):
Thank you for your contribution @dr-peppe.

sindresorhus:
@dr-peppe Bump :)

dr-peppe:
> @dr-peppe Bump :)

Sorry, I kind of forgot this as we implemented the check above as a preprocessing step for our zlib purposes for now, doing that check first before falling back to file-type for other file types. But would be nice to get rid of that and just rely on file-type of course :)

I had some problems creating files for test purposes with these other documented variations, but after looking at the Zlib API in later versions of Node I think it might be possible. I need to look into this.

sindresorhus:
@dr-peppe Still interested or should I close this?

sophacola (approved):

Issue #537, opened 2022-03-15
Issue with checking epub mimetype


prernagp:
Some epub files are not extracting mime-type properly. Getting 'application/epub+zip\r\n' when trying to check for epub files.

Some of the links for the epub files are here:-
[https://soundplay-reports.s3.amazonaws.com/1.epub](url)
[https://soundplay-reports.s3.amazonaws.com/2.epub](url)
[https://soundplay-reports.s3.amazonaws.com/3.epub](url)

The extra characters should be truncated by the library.

prernagp:
@sindresorhus Please check this issue

sindresorhus:
You can see in the source that there's no `\r\n` being added, so I don't think the issue is with `file-type`: https://github.com/sindresorhus/file-type/blob/330e6b9689ab525d71129e2cb434f91c2f18574a/core.js#L340

I'm happy to reopen if you submit a failing test.

prernagp:
@sindresorhus It was the issue with token-types library. We have raised the issue there with the solution

[https://github.com/Borewit/token-types/issues/455](https://github.com/Borewit/token-types/issues/455
)

Borewit:
@prernagp is talking about this part, where a decoding issue is (caused by the extra `\n\r` in mimeType):
https://github.com/sindresorhus/file-type/blob/cffcf1f254e3afffc91f7eec376319bdff18fd9e/core.js#L334-L336

That results in a detection of `application/zip` instead the expected `application/epub+zip`.

prernagp:
Thank you @Borewit for the help

PR #546, opened 2022-05-31
Improve decoding of  mime-type in ZIP file

Borewit:
Strip CR (Carriage Return) & LF (Line Feed) from mimeType in ZIP file decoder

Resolves sindresorhus/file-type#537


prernagp:
Thanx @sindresorhus and @Borewit

prernagp:
When this change will be released? @sindresorhus

PR #554, opened 2022-08-17
Normalized `strtok3` imports

YXL76:
Some package managers cannot resolve `strtok3/core`.


sindresorhus:
Those package managers need to be fixed: https://nodejs.org/api/packages.html#subpath-exports

Dakuan:
running into this on RN on expo, seems a shame

PR #564, opened 2022-11-08
ci: harden workflow

Fdawgs:
This PR:

-   Removes Git credentials/SSH keys after checkout as a security precaution by setting `persist-credentials` to false, they are not used after the initial checkout
- Declares the minimum permissions for CI workflow to run at the job level, following principle of least privilege; see [related GitHub security post](https://securitylab.github.com/research/github-actions-building-blocks/)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,8 @@ jobs:
   test:
     name: Node.js ${{ matrix.node-version }}
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -15,6 +17,8 @@ jobs:
           - 14
     steps:
       - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
```
sindresorhus, line 21: I don't think millions of repos should have to set this manually. The default should be changed in https://github.com/actions/checkout/pull/946

.github/workflows/main.yml
```diff
@@ -6,6 +6,8 @@ jobs:
   test:
     name: Node.js ${{ matrix.node-version }}
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
sindresorhus, line 10: This can be set for the entire workflow, not just this job.

Fdawgs:
Thanks for the feedback @sindresorhus, no worries!

Issue #579, opened 2023-02-09
End of stream error when checking the file type of a pdf


eric-yuan-vanta:
I was using an older version (16.x) and ran into this error. I then upgraded to latest, but ran into the same thing. Using the latest version (18.2.0), I've been getting the following error trace in my app with this pdf. Any idea what could be the cause?

```
{"name":"Error","stack":"Error: End-Of-Stream\n    at ReadStreamTokenizer.readBuffer (file:///node_modules/strtok3/lib/ReadStreamTokenizer.js:38:19)\n 
  at ReadStreamTokenizer.ignore (file:///node_modules/strtok3/lib/ReadStreamTokenizer.js:86:31)\n    at FileTypeParser.parse (file:///node_modules/file-type/core.js:614:4)\n    at Module.fileTypeFromStream (file:///node_modules/file-type/core.js:16:10)\n
```

I'm suspicious something is different about this pdf. I'm still able to view it with normal viewers though.

I guess we're reaching EOF somehow in the adobe ai check `await tokenizer.ignore(1350);`. Is this file missing something, or too short? Appreciate any help, thanks!

[ET-test-011222-1027-434.pdf](https://github.com/sindresorhus/file-type/files/10701372/ET-test-011222-1027-434.pdf)

eric-yuan-vanta:
I downgraded to 14.0.0, prior to the adobe illustrator support (14.1.0), and it works now.

PR #580, opened 2023-02-10
Handle very small pdf's

eric-yuan-vanta:
Handle tiny pdf's, and add a test pdf < 1350 bytes (the test pdf is MIT licensed and from https://brendanzagaeski.appspot.com/0004.html)

just fyi generating an empty pdf from google drive results in a pdf of similar size (~700 bytes)

resolves https://github.com/sindresorhus/file-type/issues/579

### DIFF
--- a/core.js
+++ b/core.js
@@ -611,17 +611,24 @@ class FileTypeParser {
 		}
 
 		if (this.checkString('%PDF')) {
-			await tokenizer.ignore(1350);
-			const maxBufferSize = 10 * 1024 * 1024;
-			const buffer = Buffer.alloc(Math.min(maxBufferSize, tokenizer.fileInfo.size));
-			await tokenizer.readBuffer(buffer, {mayBeLess: true});
+			try {
+				await tokenizer.ignore(1350);
+				const maxBufferSize = 10 * 1024 * 1024;
+				const buffer = Buffer.alloc(Math.min(maxBufferSize, tokenizer.fileInfo.size));
+				await tokenizer.readBuffer(buffer, {mayBeLess: true});
 
-			// Check if this is an Adobe Illustrator file
-			if (buffer.includes(Buffer.from('AIPrivateData'))) {
-				return {
-					ext: 'ai',
-					mime: 'application/postscript',
-				};
+				// Check if this is an Adobe Illustrator file
+				if (buffer.includes(Buffer.from('AIPrivateData'))) {
+					return {
+						ext: 'ai',
+						mime: 'application/postscript',
+					};
+				}
+			} catch (error) {
+				// Swallow end of stream error if file is too small for the Adobe AI check
+				if (!(error instanceof strtok3.EndOfStreamError)) {
+					throw error;
+				}
 			}
 
 			// Assume this is just a normal PDF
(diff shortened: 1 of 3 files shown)


eric-yuan-vanta:
Hey @sindresorhus would appreciate a review on this when you get a chance. Thanks!

core.js
```diff
@@ -611,7 +611,17 @@ class FileTypeParser {
...
-			await tokenizer.ignore(1350);
+			try {
+				await tokenizer.ignore(1350);
+			} catch (error) {
+				if (error instanceof strtok3.EndOfStreamError) {
+					return {
+						ext: 'pdf',
+						mime: 'application/pdf',
+					};
+				}
```
sindresorhus, line 622: You are swallowing other errors.

core.js
```diff
@@ -611,17 +611,24 @@ class FileTypeParser {
 		}
 
 		if (this.checkString('%PDF')) {
-			await tokenizer.ignore(1350);
-			const maxBufferSize = 10 * 1024 * 1024;
-			const buffer = Buffer.alloc(Math.min(maxBufferSize, tokenizer.fileInfo.size));
-			await tokenizer.readBuffer(buffer, {mayBeLess: true});
+			try {
+				await tokenizer.ignore(1350);
```
Borewit, line 615: I think only this line should be between the `try` and `catch`.

Maybe a more fundamental question is, why it is ignoring the first 1350 bytes if this does not represent a  portion of a PDF which is always present.

The `readBuffer` with the `mayBeLess: true` should not raise an EOF.

Borewit (changes requested):

core.js
eric-yuan-vanta: It shouldn't, but unfortunately it does

```
  fixture-minimal.pdf 295 .fromStream() method - same fileType

  test.js:300

   299:   const file = path.join(__dirname, 'fixture', filename);
   300:   const fileType = await fileTypeFromStream(fs.createReadStream(file)…
   301:

  Rejected promise returned by test. Reason:

  Error (EndOfStreamError) {
    message: 'End-Of-Stream',
  }

  › StreamReader.read (file://node_modules/peek-readable/lib/StreamReader.js:49:19)
  › ReadStreamTokenizer.readBuffer (file://node_modules/strtok3/lib/ReadStreamTokenizer.js:35:51)
  › FileTypeParser.parse (file://core.js:625:20)
  › async fileTypeFromStream (file://core.js:16:10)
  › async testFileFromStream (file://test.js:300:19)
```

failure from my test case when i only try the ignore

eric-yuan-vanta: I'm just going to early return pdf on the catch instead.

Borewit: If continue to read of the EOF is already caught, you can expect problems.

Sorry, I think you initial code was actually better. Declaring the the PDF return type twice is very nice neither. If you revert to the previous point I approve.

Borewit (changes requested):

core.js
Borewit, line 615: This remains a sensitive point, but is not introduced by this PR.
Maybe we should drop support for specialized PDF formats as text based formats are out of scope.

Borewit (approved):

PR #584, opened 2023-03-06
update import strtok3 issue fix #548

### LINKED ISSUE #548: Cannot find module 'strtok3/core' from 'node_modules/file-type/core.js'
Hello,

I have version 17.1.2 installed in a project.

When I attempt to use this library I get the error `Cannot find module 'strtok3/core' from 'node_modules/file-type/core.js'`

This seems to be caused by core.js:3 referencing `strtok3/core` instead of `strtok3/lib/core`.

Maybe this is a problem with my local setup, but wanted to see if there was an easy resolution to this issue.

abdulghani:

- **One PR per file type.**
- Add the file type detection logic to the `core.js` file
- Respect the sequence:
	- Signature with shorter sample size (counted from offset 0 until the last required byte position) will be executed first.
	- Only the initial determination for the file type counts for the sequence.
	- Existing signatures requiring same sample length (same *signature group*) will be tested prior to your new detections. Yours will be last. (rational: common formats first).
- Add the file extension to the `Supported file types` section of the readme in alphabetical order, in the format ```- [`<extension>`](URL) - Format name```, for example, ```- [`png`](https://en.wikipedia.org/wiki/Portable_Network_Graphics) - Portable Network Graphics```

### DIFF
--- a/core.js
+++ b/core.js
@@ -1,6 +1,6 @@
 import {Buffer} from 'node:buffer';
 import * as Token from 'token-types';
-import * as strtok3 from 'strtok3/core'; // eslint-disable-line n/file-extension-in-import
+import * as strtok3 from 'strtok3/lib/core'; // eslint-disable-line n/file-extension-in-import
 import {
 	stringToBytes,
 	tarHeaderChecksumMatches,


sindresorhus:
The existing code is correct. This change is not.

Issue #581, opened 2023-02-11
Importing fileTypeFromBlob in Node environment


schwartzmj:
`Blob` has been supported in Node for a bit now as I'm sure you're aware. Many projects these days develop in Node but use Edge runtimes in production.

I cannot for the life of me figure out how to configure my Vite project to allow me to import the `fileTypeFromBlob` function. `file-type` has the `exports` key in `package.json` configured as such:

```
"exports": {
	".": {
		"node": "./index.js",
		"default": "./browser.js"
	},
	"./core": "./core.js"
},
```

Anybody have a workaround? Any plans to let us just import this function in Node since `Blob` is supported?

PR #588, opened 2023-03-29
Support for reading from Blob in Node.js

Borewit:
Support for reading from [Blob](https://nodejs.org/api/buffer.html#class-blob) in Node.js context.

Resolves: #581

First merge: 
- #587

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -198,8 +198,6 @@ A readable stream representing file data.
 
 Detect the file type of a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
 
-**Note:** This method is only available in the browser.
-
 The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
 
 Returns a `Promise` for an object with the detected file type and MIME type:
--- a/browser.d.ts
+++ b/browser.d.ts
@@ -18,28 +18,9 @@ console.log(fileType);
 */
 export declare function fileTypeFromStream(stream: ReadableStream): Promise<FileTypeResult | undefined>;
 
-/**
-Detect the file type of a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
-
-__Note:__ This method is only available in the browser.
-
-@example
-```
-import {fileTypeFromBlob} from 'file-type';
-
-const blob = new Blob(['<?xml version="1.0" encoding="ISO-8859-1" ?>'], {
-	type: 'plain/text',
-	endings: 'native'
-});
-
-console.log(await fileTypeFromBlob(blob));
-//=> {ext: 'txt', mime: 'plain/text'}
-```
-*/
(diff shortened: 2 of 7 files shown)


readme.md
```diff
@@ -198,8 +198,6 @@ A readable stream representing file data.
 
 Detect the file type of a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
 
-**Note:** This method is only available in the browser.
```
sindresorhus, line 201: You need to update the TS doc comment too.

sindresorhus:
@Borewit Thanks. Maybe we can get rid of `browser.js` completely by moving the ReadableStream method into core and exposing it in Node.js if Node.js 18 (that's when it's a global) or later?

Borewit:
The current implementation `fileTypeFromStream()` converts a [Web-API readable stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader) into a [Node.js readable stream](https://nodejs.org/api/stream.html#stream_readable_streams), so essentially an [adapter](https://refactoring.guru/design-patterns/adapter).

The fact that it implemented as an [adapter](https://refactoring.guru/design-patterns/adapter) makes it unsuitable at the moment to be moved into core.

Only when we are able utilize the [Node.js ReadableStream](https://nodejs.org/api/webstreams.html#class-readablestream) it does make sense to move it into core. And actually only when this could be used to process both the 
1. [Web-API readable stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader) and the 
2. [Node.js ReadableStream](https://nodejs.org/api/webstreams.html#class-readablestream)

which should be in theory the same streams [WHATWG streams](https://streams.spec.whatwg.org/), but are in reality, not necessary interchangeable:
1. There are [issues with different TypeScript](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/55293) definitions
2. If I remember well, also the implementation was done differently hence the typings could not be blended.

**Caveat on the file-type Blob function**  

The way the implementation is, and was, inconsistent with the other function, as it does **not** utilize the lazy reading we use with streams. As it converts the Blob to a buffer, use cases where a Blob (File) is still stream (like when you upload a file), the current mechanism will read the entire Blob (File) in memory. 

Related to the previous discussion, utilizing the [Blob.stream()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream) this could be prevented. At the other hand, the user can do this as well, and provide that WHATWG stream read function we  just discussed.

jimmywarting:
if the hole file dose not need to be read in order to get magic numbers (head and footer)
then i suggest that you also slice the file
```js
head = blob.slice(0, 1024)
footer = blob.slice(-1024)
```
you could also make a really small blob out of it: `new Blob([head, footer]).stream()`

Also worth mentioning that there now also is a `fs.openAsBlob(path)` method that could make reading files (partially) easier and more cross platform ergonomic. just using `openAsBlob` alone won't allocate the file into memory. it's read lazzily when you use any of the read methods...

PR #592, opened 2023-03-31
Add support for ACE

dcfreire:
Magic number: https://en.wikipedia.org/wiki/List_of_file_signatures
MIME: https://en.wikipedia.org/wiki/List_of_archive_formats

ACE is an archiving format from 1999, it is not widely used, and the software that is capable of creating .ace files (WinAce) is abandonware. However, as per the Wikipedia article: "Since at least 2015, ACE archives have been used to deliver malware to victims by e-mail. This tactic was viable because popular archiving software was able to uncompress ACE archives, but support for the ACE format in security products (...) was generally weak.". 

Do note that I edited the magic number offset on that Wikipedia article by experimenting creating ace files on my own, with different contents, the offset was always 7 bytes, but since there is no detailed documentation on the format I cannot be sure that it is the case that it will always be the same. Still I think being able to detect it has value for the aforementioned reason.


sindresorhus:
Merge conflict

PR #597, opened 2023-04-27
avro file type support added

IgorWiecz:
Avro file type support added, steps in contributing.md file followed.


core.d.ts
```diff
@@ -167,6 +168,7 @@ export type MimeType =
 	| 'image/vnd.ms-photo'
 	| 'image/vnd.adobe.photoshop'
 	| 'application/x-indesign'
+	| 'application/avro'
```
sindresorhus, line 171: Should be added to the end.

core.d.ts
```diff
@@ -2,6 +2,7 @@ import type {Readable as ReadableStream} from 'node:stream';
 import type {ITokenizer} from 'strtok3';
 
 export type FileExtension =
+	| 'avro'
```
sindresorhus, line 5: Should be added to the end.

IgorWiecz: Done.

core.d.ts
IgorWiecz: Done.

core.js
```diff
@@ -292,6 +299,13 @@ class FileTypeParser {
 			};
 		}
 
+		if (this.checkString('avro')) {
```
sindresorhus, line 302: Can you link to a source for this?

supported.js
```diff
@@ -1,4 +1,5 @@
 export const extensions = [
+	'avro',
```
sindresorhus, line 2: It was implied that https://github.com/sindresorhus/file-type/pull/597#discussion_r1179941635 applied here too. And the below.

core.js
sindresorhus: I don't see it written anywhere there that the file header must start with `avro`.

IgorWiecz: I've also noticed people use .avsc and .acme, but I don't quite understand the difference

PR #601, opened 2023-05-31
Add support for icc profile

IgorWiecz:

- **One PR per file type.**
- Add the file type detection logic to the `core.js` file
- Respect the sequence:
	- Signature with shorter sample size (counted from offset 0 until the last required byte position) will be executed first.
	- Only the initial determination for the file type counts for the sequence.
	- Existing signatures requiring same sample length (same *signature group*) will be tested prior to your new detections. Yours will be last. (rational: common formats first).
- Add the file extension to the `Supported file types` section of the readme in alphabetical order, in the format ```- [`<extension>`](URL) - Format name```, for example, ```- [`png`](https://en.wikipedia.org/wiki/Portable_Network_Graphics) - Portable Network Graphics```


core.js
```diff
@@ -1311,6 +1311,13 @@ class FileTypeParser {
 		// Increase sample size from 12 to 256.
 		await tokenizer.peekBuffer(this.buffer, {length: Math.min(256, tokenizer.fileInfo.size), mayBeLess: true});
 
+		if (this.check([0x61, 0x63, 0x73, 0x70], {offset: 36})) {
```
sindresorhus, line 1314: Can you link to a source for this magic number?

PR #603, opened 2023-07-17
Extend file type (updated)

### LINKED ISSUE #340: Extending file-type

FredrikSchaefer:
Hey There :)

I came across this issue #340 and found that it's just what we need. As the [initial PR](https://github.com/sindresorhus/file-type/pull/399) is way outdated, I decided to start over from scratch.

Resolves #340.


FredrikSchaefer:
Gentle nudge towards @sindresorhus . We are really looking forward to using this feature in the next official version.

sindresorhus:
I'm on vacation, so I won't be able to review this for another couple of weeks.

core.d.ts
```diff
@@ -318,6 +318,38 @@ export type ReadableStreamWithFileType = ReadableStream & {
...
+			}
+
+			return undefined;
+		},
+	];
+
+ @param tokenizer - An [`ITokenizer`](https://github.com/Borewit/strtok3#tokenizer) usable as source of the examined file.
+ @param fileType - FileTypeResult detected by the standard detections or a previous custom detection. Undefined if no matching fileTypeResult could be found.
+ @returns supposedly detected file extension and MIME type as a FileTypeResult-like object, or `undefined` when there is no match.
+ */
```
sindresorhus, line 350:
```suggestion
*/
```

core.d.ts
```diff
@@ -326,19 +358,21 @@ The file type is detected by checking the [magic number](https://en.wikipedia.or
 If file access is available, it is recommended to use `.fromFile()` instead.
 
 @param buffer - An Uint8Array or Buffer representing file data. It works best if the buffer contains the entire file, it may work with a smaller portion as well.
+@param customDetectors - Optional: An Iterable of Detector functions. They are called in the order provided.
 @returns The detected file type and MIME type, or `undefined` when there is no match.
 */
-export function fileTypeFromBuffer(buffer: Uint8Array | ArrayBuffer): Promise<FileTypeResult | undefined>;
+export function fileTypeFromBuffer(buffer: Uint8Array | ArrayBuffer, customDetectors?: Iterable<Detector>): Promise<FileTypeResult | undefined>;
```
sindresorhus, line 364: `customDetectors` should be an option in an options-object, not a separate parameter.

sindresorhus (changes requested):

FredrikSchaefer:
> I'm on vacation, so I won't be able to review this for another couple of weeks.

Thanks a lot for squeezing in time to send change requests, even on vacation! We will then see that we don't rely on quick action from your side. Enjoy your time off!

Borewit:
I am on holiday as well, this time without a laptop. Back in 2 weeks.

FredrikSchaefer:
Another gentle nudge towards @sindresorhus and @Borewit , so this PR does not get lost in the the vortex of time.

FredrikSchaefer:
... And yet another gentle nudge to @sindresorhus . I'd really appreciate if you found the time for a review :-)

FredrikSchaefer:
Is there anything I can do to support the review of this PR?

Borewit:
I will try find some time as well to review this PR @FredrikSchaefer. Please give me 2 weeks.

core.d.ts
```diff
@@ -318,6 +318,42 @@ export type ReadableStreamWithFileType = ReadableStream & {
...
+/**
+Function that allows specifying custom detection mechanisms.
+
+An iterable of detectors can be provided as argument for file type detection options.
+
+The detectors are called before the default detections in the provided order.
+
+Custom detectors can be used to add new FileTypeResults or to modify return behaviour of existing FileTypeResult detections.
+
+If the detector returns `undefined`, the `tokenizer.position` should be 0 (unless it's a stream). That allows other detectors to parse the file.
```
Borewit, line 330: Not sure if I agree with the "unless it's a stream". Essentially you can iterate to other detectors if you took _a bite of the apple_. Only peek is allowed, if read you you have consumed the tokenizer, which is very similar to a stream. 

I fear this an area where we can expect a lot questions from users.

core.d.ts
```diff
@@ -326,19 +362,21 @@ The file type is detected by checking the [magic number](https://en.wikipedia.or
 If file access is available, it is recommended to use `.fromFile()` instead.
 
 @param buffer - An Uint8Array or Buffer representing file data. It works best if the buffer contains the entire file, it may work with a smaller portion as well.
+@param fileTypeOptions - Optional: An options object including the `customDetectors` property as an Iterable of Detector functions. Those are called in the order provided.
 @returns The detected file type and MIME type, or `undefined` when there is no match.
 */
-export function fileTypeFromBuffer(buffer: Uint8Array | ArrayBuffer): Promise<FileTypeResult | undefined>;
+export function fileTypeFromBuffer(buffer: Uint8Array | ArrayBuffer, fileTypeOptions?: FileTypeOptions): Promise<FileTypeResult | undefined>;
```
Borewit, line 368: Maybe a bit drastic, but I am actually in favor introducing a constructor. To me it does not make sense to initialize the scanner in each function.

Borewit (changes requested):
Thank you so much for your patience @FredrikSchaefer. 

I think you have done a fantastic job. I just have some minor comment.

core.d.ts
FredrikSchaefer: Hey, thanks for the comment.

Yeah, guess you're right about the lot of questions.

I just mindlessly took this information from [this previous discussion](https://github.com/sindresorhus/file-type/pull/399/files#r496119460).

Let me suggest a more detailed explanation here:

If the detector returns `undefined`, the `tokenizer.position` should typically be 0. This allows easy parsing by other detectors, unless subsequent custom detectors specify otherwise. Additionally, the detector shouldn't consume the tokenizer; while peeking is non-consuming, reading is.

What do you think of this?

I'm really open to anything here!

core.d.ts
FredrikSchaefer: Let's discuss that a bit. I argue against it because:

- I agree that it's a bit drastic. If I get you correctly, that would be a breaking change, just for this one feature. I guess many users prefer not needing to change anything over having the constructor for this one additional argument?
- Two arguments for a function is not so much. @sindresorhus already brought up the use of a fileTypeOptions Object here, which is extendable and therefore rather future-proof: Additional options can just be put into that and don't need to add additional parameters to the functions.

Would love to hear your thoughts on this.

core.d.ts
```diff
@@ -318,6 +318,42 @@ export type ReadableStreamWithFileType = ReadableStream & {
...
+	const customDetectors = [
+		async tokenizer => {
+			const unicornHeader = [85, 78, 73, 67, 79, 82, 78]; // "UNICORN" as decimal string
+			const buffer = Buffer.alloc(7);
+			await tokenizer.peekBuffer(buffer, {length: unicornHeader.length, mayBeLess: true});
+			if (unicornHeader.every((value, index) => value === buffer[index])) {
+				return {ext: 'unicorn', mime: 'application/unicorn'};
+			}
+
+			return undefined;
```
Borewit, line 343: I just realized, to give the custom scanners the same power as the internal scanner, you may want to be able to distinct the following cases:
1. Scanner does recognize the file, it did not read from the tokenizer, proceed with the next scanner
2. Scanner does kind if recognize the file, but the file exact type could not be determined, hence it also returns undefined. Possibly the scanner went in "level 2 scaniing" and we (read) consumed the data.

Examples of option 2:
- https://github.com/sindresorhus/file-type/blob/ca7a03c26761444e024764a38d16597455281cfd/core.js#L1267
- https://github.com/sindresorhus/file-type/blob/ca7a03c26761444e024764a38d16597455281cfd/core.js#L1483

core.d.ts
Borewit: See also my other comment: https://github.com/sindresorhus/file-type/pull/603/files#r1356979704

I suggest something like this:

If the detector returns _no_match_ it is not allowed to read from the tokenizer (the tokeinzer.position must remain 0) otherwise following scanners will read from the wrong file offset.

If the detector return _undefined_ the scanner is certain the file type cannot be determined, neither by other scanners.

_no_match_ represents option 1 explained in [here](https://github.com/sindresorhus/file-type/pull/603/files#r1356979704)

FredrikSchaefer: I agree with your point that custom detectors should be able to interrupt detection. However, I see two small downsides of the suggested approach:

1. The wording _no_match_ does not really make clear to me whether it means no match **for this detector,** or no match **at all**.

2. The standard FileTypeParser returns undefined when no file type could be recognized. Therefore requiring the custom detectors to return something else is a bit counter intuitive.

I therefore suggest to do it the other way around:

If the detector returns _undefined_, it is not allowed to read from the tokenizer (the tokenizer.position must remain 0) otherwise following scanners will read from the wrong file offset.

If the detector returns _file_type_undetectable_, the detector is certain the file type cannot be determined, even by other scanners. The FileTypeParser interrupts the parsing and immediately returns undefined.

FredrikSchaefer: Okay, one could argue that _file_type_undetectable_ also does not clearly say whether it means file type undetectable **for this detector** or **for all detectors**, but it still makes it a bit clearer in my opinion.

core.d.ts
FredrikSchaefer: Please note my reply [above](https://github.com/sindresorhus/file-type/pull/603#discussion_r1361559226).

core.d.ts
Borewit: > I agree with your point that custom detectors should be able to interrupt detection. However, I see two small downsides of the suggested approach:
> 
> 
> 
> 1. The wording _no_match_ does not really make clear to me whether it means no match **for this detector,** or no match **at all**.
> 
> 
> 
> 2. The standard FileTypeParser returns undefined when no file type could be recognized. Therefore requiring the custom detectors to return something else is a bit counter intuitive.
> 
> 
> 
> I therefore suggest to do it the other way around:
> 
> 
> 
> If the detector returns _undefined_, it is not allowed to read from the tokenizer (the tokenizer.position must remain 0) otherwise following scanners will read from the wrong file offset.
> 
> 
> 
> If the detector returns _file_type_undetectable_, the detector is certain the file type cannot be determined, even by other scanners. The FileTypeParser interrupts the parsing and immediately returns undefined.

Sounds good to me. The second case can be also be something like, detector started reading but for some reason failed to determine the file-type. Not ideally, but it can happen. If the detector starts reading, there is no way back.

We could also check the position after each custom scanner. It may not be 0 actually, there is also an iterated use case with ID3 header. The position should be remain unchanged.

FredrikSchaefer: Good idea! Just pushed a commit taking care of that check.

core.js
```diff
@@ -1602,7 +1662,7 @@ class FileTypeParser {
 	}
 }
 
-export async function fileTypeStream(readableStream, {sampleSize = minimumBytes} = {}) {
+export async function fileTypeStream(readableStream, {sampleSize = minimumBytes} = {}, fileTypeOptions) {
```
Borewit, line 1665: Inconsistent with TypeScript declaration. 

Why is the class constructor not used to initialize with the `fileTypeOptions`?

core.js
```diff
@@ -60,16 +44,92 @@ function _check(buffer, headers, options) {
...
+
+		return this.fileTypeFromTokenizer(strtok3.fromBuffer(buffer));
+	}
+
+	async fileTypeFromBlob(blob) {
+		const buffer = await blob.arrayBuffer();
+		return this.fileTypeFromBuffer(new Uint8Array(buffer));
+	}
+
+	async fileTypeFromStream(stream) {
```
Borewit, line 124: Inconsistent with function

core.d.ts
```diff
@@ -435,3 +439,91 @@ console.log(await fileTypeFromBlob(blob));
...
+		},
+	];
```
Borewit, line 470:
```suggestion
```js
const customDetectors = [
	async tokenizer => {
		const unicornHeader = [85, 78, 73, 67, 79, 82, 78]; // "UNICORN" as decimal string
		const buffer = Buffer.alloc(7);
		await tokenizer.peekBuffer(buffer, {length: unicornHeader.length, mayBeLess: true});
		if (unicornHeader.every((value, index) => value === buffer[index])) {
			return {ext: 'unicorn', mime: 'application/unicorn'};
		}
		return undefined;
	}
]
```
```

File: core.d.ts [hunk 9]
```diff
@@ -435,3 +439,91 @@ console.log(await fileTypeFromBlob(blob));
...
+
+	Example usage:
+
+	const buffer = ...
+	const parser = new FileTypeParser({customDetectors});
+	const fileType = await parser.fileTypeFromBuffer(buffer);
```
Borewit, line 476:
```suggestion
```js
const buffer = ...
const parser = new FileTypeParser({customDetectors});
const fileType = await parser.fileTypeFromBuffer(buffer);
```
```

File: readme.md [hunk 10]
```diff
@@ -469,6 +480,61 @@ The following file types will not be accepted:
 - `.csv` - [Reason.](https://github.com/sindresorhus/file-type/issues/264#issuecomment-568439196)
 - `.svg` - Detecting it requires a full-blown parser. Check out [`is-svg`](https://github.com/sindresorhus/is-svg) for something that mostly works.
 
+## Custom detectors
```
Borewit, line 483: I suggest to move it before [_Supported File Types_](https://github.com/FredrikSchaefer/file-type/tree/extend-file-type-updated#supported-file-types), so that the API documentation is not fragmented.

Borewit (changes requested):
I don't think we should throw an error when the scanner has moved the position and returned null. As it may not necessary bad implementation of the custom scanner. 

We just have to return undefined (so not execute any of the following scanners).

readme.md
FredrikSchaefer: Done.

core.js
FredrikSchaefer: Well, from the code structure, this method seemed a bit distinct (being separated from the others). I also found it a bit more intuitive that the class constructor is not used here, to better indicate that the stream is not stored.

I don't have a strong inclination here, so we could also use the class constructor. In case you don't agree to my reasoning?

core.js
FredrikSchaefer: I don't see what you mean, I'm afraid. Do you refer to `this.fileTypeFromTokenizer(tokenizer);` being called instead of `fileTypeFromTokenizer(tokenizer);`?

FredrikSchaefer:
> I don't think we should throw an error when the scanner has moved the position and returned null. As it may not necessary bad implementation of the custom scanner.
> 
> We just have to return undefined (so not execute any of the following scanners).

Wouldn't that be a bit inconsistent with the wording [here](https://github.com/sindresorhus/file-type/pull/603#discussion_r1362340706)?

It currently says:

> If the detector returns undefined, it is not allowed to read from the tokenizer (the tokenizer.position must remain 0) otherwise following scanners will read from the wrong file offset.
> 
> If the detector returns file_type_undetectable, the detector is certain the file type cannot be determined, even by other scanners. The FileTypeParser interrupts the parsing and immediately returns undefined.

This indicates to me that, in the first scenario the logic behaves different from the second scenario, and also `it is not allowed` and `must` tells me that doing so is really undesired. I think it would help debugging if an error results out of that.

If you don't agree, should we adjust the wording?

core.js
Borewit: My bad, my comment was applicable to the last line only:

```js
async fileTypeFromStream(stream) {
```
which is different then:
```
fileTypeStream(readableStream, {sampleSize = minimumBytes} = {}, fileTypeOptions)
```
the sample size was probable already missing.

So it's essentially inconsistent with TypeScript declaration, as reported in https://github.com/sindresorhus/file-type/pull/603#discussion_r1367942822

core.js
Borewit: I agree a it is a bit different indeed, but don't see a good reason to treat it any different.

core.js
```diff
@@ -60,16 +44,92 @@ function _check(buffer, headers, options) {
...
+
+			if (initialPosition !== tokenizer?.position) {
+				throw new TokenizerPositionError();
+			}
+		}
+
+		return undefined;
+	}
+
+	async fileTypeFromTokenizer(tokenizer) {
```
Borewit, line 93: Member function of the class do not require the `fileType` prefix, which was introduced when we switched to ESM. As in the CJS the functions were typically used in combination with the module reference

```suggestion
	async fromTokenizer(tokenizer) {
```

core.js
```diff
@@ -60,16 +44,92 @@ function _check(buffer, headers, options) {
...
+			const customFileType = await runCustomDetectors(tokenizer);
+			return (customFileType || await parse(tokenizer));
+		} catch (error) {
+			if (!(error instanceof strtok3.EndOfStreamError)) {
+				throw error;
+			}
+		}
+	}
+
+	async fileTypeFromBuffer(input) {
```
Borewit, line 105:
```suggestion
	async fomBuffer(input) {
```

core.js
```diff
@@ -60,16 +44,92 @@ function _check(buffer, headers, options) {
...
+	constructor(options) {
+		this.detectors = options?.customDetectors;
+
+		this.runCustomDetectors = this.runCustomDetectors.bind(this);
+		this.fileTypeFromTokenizer = this.fileTypeFromTokenizer.bind(this);
+		this.fileTypeFromBuffer = this.fileTypeFromBuffer.bind(this);
+		this.parse = this.parse.bind(this);
+	}
+
+	async runCustomDetectors(tokenizer) {
```
Borewit, line 75: Maybe make this one private?

```suggestion
	async #runCustomDetectors(tokenizer) {
```

core.js
```diff
@@ -60,16 +44,92 @@ function _check(buffer, headers, options) {
...
+	 * when they are certain that the file type can not be determined,
+	 * neither by the detector itself, nor by any further detector.
+	 * This will interrupt the file type detection.
+	 */
+	static detectionImpossible = 'detection_impossible';
+
+	constructor(options) {
+		this.detectors = options?.customDetectors;
+
+		this.runCustomDetectors = this.runCustomDetectors.bind(this);
```
Borewit, line 69: Maybe a stupid question, why do we need the binds?

Borewit:
> If you don't agree, should we adjust the wording?

What I am saying is, and what I have been saying, is that when a detector decides to to a deep scan (start reading / consuming the stream / tokenizer) it may not necessary return a file type, but return undefined instead. In that case file-type should stop trying other detectors as the stream has been consumed and return undefined as the final conclusion.

For example when you a detector detects an ID3 header, which can be 2 MB, you need to read from the stream to get to the actual audio part. There is no guarantee the detector will be able to detect the file-type coming after that. So let's assume this is the case, there is just white noise after the ID3 header.
In that case we cannot fall back on any more generic file-type, as ID3 header is insufficient to determine the file-type, highest probability it is audio/mpeg, but would be speculation. The detector should return undefined here, the position has been moved so we do not proceed with any other detectors, we should return undefined to the user. There is no reason to raise an error.

Borewit:
Before someone else is pointing this out, _so why are we guessing the file-type then in the existing code?_

https://github.com/sindresorhus/file-type/blob/ca7a03c26761444e024764a38d16597455281cfd/core.js#L206-L210

Good question.

core.js
FredrikSchaefer: I don't think it's a stupid questions, I stumbled upon this myself :)
These bindings are necessary to prevent `this.[...]` from being lost in the `fromTokenizer` and `parse` method.

FredrikSchaefer:
> > If you don't agree, should we adjust the wording?
> 
> What I am saying is, and what I have been saying, is that when a detector decides to to a deep scan (start reading / consuming the stream / tokenizer) it may not necessary return a file type, but return undefined instead. In that case file-type should stop trying other detectors as the stream has been consumed and return undefined as the final conclusion.
> 
> For example when you a detector detects an ID3 header, which can be 2 MB, you need to read from the stream to get to the actual audio part. There is no guarantee the detector will be able to detect the file-type coming after that. So let's assume this is the case, there is just white noise after the ID3 header. In that case we cannot fall back on any more generic file-type, as ID3 header is insufficient to determine the file-type, highest probability it is audio/mpeg, but would be speculation. The detector should return undefined here, the position has been moved so we do not proceed with any other detectors, we should return undefined to the user. There is no reason to raise an error.

Thank you very much for the detailed explanation. It helps me understand your reasoning.
In that scenario, doesn't it make more sense to use recursion? 
Example:

```
let parser;
const recursiveDetector = async tokenizer => {

// Some logic reading from the tokenizer

await parser.fromTokenizer(tokenizer);
};

const customDetectors = [recursiveDetector];
parser = new FileTypeParser({customDetectors});

const result = await parser.fromTokenizer(someTokenizer);
```

All in all, the current behavior looks like being a bit clearer and less prone to error, while still being flexible enough to not block any valid use.
In case you insist on that change, I'd really appreciate if you kindly provided an updated version of the docstring [here](https://github.com/sindresorhus/file-type/pull/603#discussion_r1362340706). Honestly, I would like to finalize this soon, and therefore it would be great if we could cut down the number of review iterations a bit...

core.js
FredrikSchaefer: Adjusted.

core.js
FredrikSchaefer: Aren't `fromStream(...)` and `stream(...)` methods getting mixed up here? (`fileTypeFromStream(...)` and `fileTypeStream(...)` respectively)

Borewit:
> I would like to finalize this soon

Sure, don't throw the error. Yes we do use recursion in that scenario, but that does change anything, it help to make that possible.

> I'd really appreciate if you kindly provided an updated version of the docstring

I propose something like this:

Custom detectors can be used to add new FileTypeResults or to modify return behaviour of existing FileTypeResult detections.
If the detector returns `undefined`, there are 2 possible scenario's:
1. The detector has not read from the tokenizer, it will proceed with the next available detector 
2. The detector has read from the tokenizer (tokenzier.position is has not been increased), in that case no further detectors will be executed and the final conclusion is that file-type returns `undefined`. Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.

FredrikSchaefer:
I'm sorry, I missed that part in explaining my reasoning:

In the case of recursion (following my argumentation), the logic should look like this:

```
let parser;

const recursiveDetector = async tokenizer => {

	// Some logic reading from the tokenizer

	return await parser.fromTokenizer(tokenizer) || FileTypeParser.detectionImpossible;
};

const customDetectors = [recursiveDetector];
parser = new FileTypeParser({customDetectors});

const result = await parser.fromTokenizer(someTokenizer);
```

If that still doesn't convince you, I'll not throw the error.

Borewit:
We essentially agreed to replace the explicit "detection-impossible" with a mechanism that detects the tokenizerer has been consumed (result == undefined & tokenizer-position-has-been-changed). So that's the same thing, but easier to implement.

Yes, it is a risk (limitation) users will read from the tokenizer, and are unable to detect the type, therefore return null, and misunderstand the mechanism and expect the remaining detectors to proceed. But if we throw an error, we also make perfectly valid cases impossible to implement. 
 
That same risks (limitation) is present in the internal detectors. If we restrict the power of the custom detectors they are pretty much useless, as there is mostly a need to cover complex detection (require deep reading) more then simple (first few bytes), since most of those simple ones are probably already implemented by us.

If I borrow you my car, you can crash it.

FredrikSchaefer:
Okay then. I adjusted the logic. The [docstring you suggested](https://github.com/sindresorhus/file-type/pull/603#issuecomment-1776856930) is also included, I just polished it a bit.

core.js
```diff
@@ -60,16 +44,120 @@ function _check(buffer, headers, options) {
...
+		const initialPosition = tokenizer?.position;
+
+		for (const detector of detectors || []) {
+			const fileType = await detector(tokenizer);
+			if (fileType) {
+				return fileType;
+			}
+
+			if (initialPosition !== tokenizer?.position) {
+				throw new TokenizerPositionError();
```
midmarch, line 78: As discussed
```suggestion
				return undefined;
```

core.js
```diff
@@ -60,16 +44,120 @@ function _check(buffer, headers, options) {
...
+			}
+		}
+
+		return undefined;
+	}
+
+	async fromTokenizer(tokenizer) {
+		const {parse, runCustomDetectors} = this;
+		try {
+			const customFileType = await runCustomDetectors(tokenizer);
```
midmarch, line 88: You miss the check here if the tokenizer has been read.

You could consider to call parse from `#runCustomDetectors(tokenizer)`, after the for loop

core.js
```diff
@@ -60,16 +44,120 @@ function _check(buffer, headers, options) {
...
+	async fromStream(stream) {
+		const tokenizer = await strtok3.fromStream(stream);
+		try {
+			return await this.fromTokenizer(tokenizer);
+		} finally {
+			await tokenizer.close();
+		}
+	}
+
+	async stream(readableStream, options = {}) {
```
midmarch, line 125: Suggestion, maybe good to articulate the different functionality here.
```suggestion
	async toDetectingStream(readableStream, options = {}) {
```

core.d.ts
```diff
@@ -435,3 +438,93 @@ console.log(await fileTypeFromBlob(blob));
...
+	 *
+	 * Works the same way as {@link fileTypeFromBlob}, additionally taking into account custom detectors (if any were provided to the constructor).
+	 */
+	fromBlob(blob: Blob): Promise<FileTypeResult | undefined>;
+
+	/**
+	 *
+	 * Works the same way as {@link fileTypeStream}, additionally taking into account custom detectors (if any were provided to the constructor).
+	 */
+	stream(readableStream: ReadableStream, options?: StreamOptions): Promise<FileTypeResult | undefined>;
```
midmarch, line 529: Suggestion, maybe good to articulate the different functionality here.
```suggestion
	toDetectingStream(readableStream: ReadableStream, options?: StreamOptions): Promise<FileTypeResult | undefined>;
```
Same comment provided for the implementation

midmarch (changes requested):
Almost there

Borewit:
Forgive me, wrong account, same person.

core.js
FredrikSchaefer: In my understanding we agreed on not throwing the error to the user.
 The error is caught in the `FileTypeParser.fromTokenizer(...)`, which blends in nicely to the existing logic.
 
```
 		} catch (error) {
			if (!(error instanceof strtok3.EndOfStreamError) && !(error instanceof TokenizerPositionError)) {
				throw error;
			}
		}
```
What's the problem with this approach?

core.js
FredrikSchaefer: I'm sorry, I don't see in what scenario the check is missed. It happens inside the `runCustomDetectors(...)`, and I think that's where it belongs, as it's part of the custom detectors functionality.

core.js
FredrikSchaefer: Done.

core.d.ts
FredrikSchaefer: Done.

core.js
midmarch: How do you prevent internal detectors from running when a custom detector has consumed the stream?

You have now protected the remaing detectors when undefined & tokenizer has been read, situation occured, but not the internal detectors.

Borewit: Sorry I missed the part that you further on were catching this error, it's something I did not directly expect.

I made some suggestions in a branch, how I believe the same thing can be achieved, simpler, with less code: https://github.com/sindresorhus/file-type/tree/extend-file-type-updated-suggested-code-changes

FredrikSchaefer: Thank you very much for making the effort. Looks okay to me, just merged it.

Borewit (approved):
Looks good to me, thanks a lot for your effort @FredrikSchaefer

midmarch (approved):
Ignore me

Borewit:
@sindresorhus, kind reminder, please proceed with merging if you are happy with this one

index.d.ts
```diff
@@ -1,13 +1,14 @@
...
 Detect the file type of a file path.
 
 The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
 
 @param path - The file path to parse.
+@param fileTypeOptions - Optional: An options object including the `customDetectors` property as an Iterable of Detector functions. Those are called in the order provided.
 @returns The detected file type and MIME type or `undefined` when there is no match.
 */
-export function fileTypeFromFile(path: string): Promise<FileTypeResult | undefined>;
+export function fileTypeFromFile(path: string, fileTypeOptions?: FileTypeOptions): Promise<FileTypeResult | undefined>;
```
sindresorhus, line 12:
```suggestion
export function fileTypeFromFile(path: string, options?: FileTypeOptions): Promise<FileTypeResult | undefined>;
```

sindresorhus: Do we really need to expose it on the methods here now that we expose the class? I thought the point of exposing the class was so we didn't have to add the option to each method.

readme.md
```diff
@@ -305,6 +309,49 @@ Returns a `Set<string>` of supported file extensions.
...
+If the detector returns `undefined`, there are 2 possible scenarios:
+
+1. The detector has not read from the tokenizer, it will be proceeded with the next available detector.
+2. The detector has read from the tokenizer (`tokenizer.position` has been increased).
+		In that case no further detectors will be executed and the final conclusion is that file-type returns undefined.
+		Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.
+
+
+Example detector array which can be extended and provided to each public method via the fileTypeOptions argument:
+```
```
sindresorhus, line 331:
```suggestion
```js
```

test.js
```diff
@@ -661,3 +663,180 @@ test('corrupt MKV throws', async t => {
...
+	const uint8ArrayContent = new TextEncoder().encode(header);
+
+	// Include the unicormDetector here to verify it's not used after the tokenizer.position changed
+	const customDetectors = [tokenizerPositionChanger, unicornDetector];
+	const parser = new FileTypeParser({customDetectors});
+
+	const result = await parser.fromTokenizer(strtok3.fromBuffer(uint8ArrayContent));
+	t.true(result === undefined);
+});
+
```
sindresorhus, line 842:
```suggestion
```

test.js
```diff
@@ -661,3 +663,180 @@ test('corrupt MKV throws', async t => {
...
+	const customDetectors = [mockPngDetector];
+	const parser = new FileTypeParser({customDetectors});
+
+	const result = await parser.fromStream(readableStream);
+	t.deepEqual(result, {ext: 'mockPng', mime: 'image/mockPng'});
+});
+
+test('fileTypeFromFile should detect custom file type "unicorn" using custom detectors', async t => {
+	// Load unicorn file content
```
sindresorhus, line 801: This line is not "loading" anything, and I don't think the comment adds anything. Same with the other tests.

core.d.ts
```diff
@@ -435,3 +438,93 @@ console.log(await fileTypeFromBlob(blob));
...
+```js
+const buffer = ...
+const parser = new FileTypeParser({customDetectors});
+const fileType = await parser.fromBuffer(buffer);
+
+	fromStream(...), fromTokenizer(...), fromBlob(...) and toDetectingStream(...) are available in the same manner.
+
+ @param tokenizer - An [`ITokenizer`](https://github.com/Borewit/strtok3#tokenizer) usable as source of the examined file.
+ @param fileType - FileTypeResult detected by the standard detections or a previous custom detection. Undefined if no matching fileTypeResult could be found.
+ @returns supposedly detected file extension and MIME type as a FileTypeResult-like object, or `undefined` when there is no match.
```
sindresorhus, line 484:
```suggestion
@returns supposedly detected file extension and MIME type as a FileTypeResult-like object, or `undefined` when there is no match.
```

readme.md
```diff
@@ -305,6 +309,49 @@ Returns a `Set<string>` of supported file extensions.
...
+		}
+
+		return undefined;
+	},
+];
+```
+
+Example usage:
+```
+const buffer = ...
```
sindresorhus, line 348: Code examples should be copy-pastable and include required code to run, including imports.

readme.md
```diff
@@ -305,6 +309,49 @@ Returns a `Set<string>` of supported file extensions.
...
+const parser = new FileTypeParser({customDetectors});
+const fileType = await parser.fromBuffer(buffer);
+// fromStream(...), fromTokenizer(...), fromBlob(...) and toDetectingStream(...) are available in the same manner.
+
```
sindresorhus, line 352:
```suggestion
```

sindresorhus (changes requested):

core.js
```diff
@@ -60,16 +44,98 @@ function _check(buffer, headers, options) {
...
+	async fromStream(stream) {
+		const tokenizer = await strtok3.fromStream(stream);
+		try {
+			return await this.fromTokenizer(tokenizer);
+		} finally {
+			await tokenizer.close();
+		}
+	}
+
+	async toDetectingStream(readableStream, options = {}) {
```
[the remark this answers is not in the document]
Borewit: > Why not just `toStream`?

Because that expresses poorly what this function is for.
It is not about creating a stream, it creates a special stream from a conventional stream, able to detect the file type.

index.d.ts
Borewit: Totally agree

core.js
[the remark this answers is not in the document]
sindresorhus: Ok. But it should be called `toDetectionStream`.

core.d.ts
```diff
@@ -419,7 +419,10 @@ if (stream2.fileType?.mime === 'image/jpeg') {
 export function fileTypeStream(readableStream: ReadableStream, options?: StreamOptions): Promise<ReadableStreamWithFileType>;
 
 /**
-Detect the file type of a [`Blob`](https://nodejs.org/api/buffer.html#class-blob).
+Detect the file type of a [`Blob`](https://nodejs.org/api/buffer.html#class-blob) or .
```
sindresorhus, line 422:
```suggestion
Detect the file type of a [`Blob`](https://nodejs.org/api/buffer.html#class-blob) or .
```

"Or" what?

core.d.ts
```diff
@@ -419,7 +419,10 @@ if (stream2.fileType?.mime === 'image/jpeg') {
...
+
+@param blob [`Blob`](https://nodejs.org/api/buffer.html#class-blob) used for file detection
```
sindresorhus, line 424: This is moot. You are restarting the type which is already made clear in the type signature, and the description doesn't add anything.

core.d.ts
```diff
@@ -435,3 +438,88 @@ console.log(await fileTypeFromBlob(blob));
...
+If the detector returns `undefined`, there are 2 possible scenarios:
+
+	1. The detector has not read from the tokenizer, it will be proceeded with the next available detector.
+	2. The detector has read from the tokenizer (`tokenizer.position` has been increased).
+		 In that case no further detectors will be executed and the final conclusion is that file-type returns undefined.
+		 Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.
+
+Example detector array which can be extended and provided via the fileTypeOptions argument:
+
+```js
```
sindresorhus, line 460:
```suggestion
```
```

File: core.d.ts [hunk 30]
```diff
@@ -435,3 +438,88 @@ console.log(await fileTypeFromBlob(blob));
...
+The detectors are called before the default detections in the provided order.
+
+Custom detectors can be used to add new FileTypeResults or to modify return behaviour of existing FileTypeResult detections.
+
+If the detector returns `undefined`, there are 2 possible scenarios:
+
+	1. The detector has not read from the tokenizer, it will be proceeded with the next available detector.
+	2. The detector has read from the tokenizer (`tokenizer.position` has been increased).
+		 In that case no further detectors will be executed and the final conclusion is that file-type returns undefined.
+		 Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.
```
sindresorhus, line 456:
```suggestion
1. The detector has not read from the tokenizer, it will be proceeded with the next available detector.
2. The detector has read from the tokenizer (`tokenizer.position` has been increased).
	In that case no further detectors will be executed and the final conclusion is that file-type returns undefined.
	Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.
```

core.d.ts
```diff
@@ -435,3 +438,88 @@ console.log(await fileTypeFromBlob(blob));
...
+	},
+];
+
+const buffer = Buffer.from("UNICORN");
+const parser = new FileTypeParser({customDetectors});
+const fileType = await parser.fromBuffer(buffer);
+console.log(fileType);
+```
+
+@param tokenizer - An [`ITokenizer`](https://github.com/Borewit/strtok3#tokenizer) usable as source of the examined file.
```
sindresorhus, line 482: There is no need to mention the type as it's already in the type signature.

core.d.ts
```diff
@@ -435,3 +438,88 @@ console.log(await fileTypeFromBlob(blob));
...
+	constructor(message?: string);
+}
+
+export declare class FileTypeParser {
+	detectors: Iterable<Detector>;
+
+	constructor(options?: {customDetectors?: Iterable<Detector>});
+
+	/**
+	 * Works the same way as {@link fileTypeFromBuffer}, additionally taking into account custom detectors (if any were provided to the constructor).
```
sindresorhus, line 502:
```suggestion
	Works the same way as {@link fileTypeFromBuffer}, additionally taking into account custom detectors (if any were provided to the constructor).
```

core.d.ts
```diff
@@ -435,3 +438,88 @@ console.log(await fileTypeFromBlob(blob));
...
+	 */
```
sindresorhus, line 503:
```suggestion
	*/
```

readme.md
```diff
@@ -305,6 +309,47 @@ Returns a `Set<string>` of supported file extensions.
...
+The detectors are called before the default detections in the provided order.
+
+Custom detectors can be used to add new `FileTypeResults` or to modify return behaviour of existing FileTypeResult detections.
+
+If the detector returns `undefined`, there are 2 possible scenarios:
+
+1. The detector has not read from the tokenizer, it will be proceeded with the next available detector.
+2. The detector has read from the tokenizer (`tokenizer.position` has been increased).
+		In that case no further detectors will be executed and the final conclusion is that file-type returns undefined.
+		Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.
```
sindresorhus, line 327:
```suggestion
	In that case no further detectors will be executed and the final conclusion is that file-type returns undefined.
	Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.
```

readme.md
```diff
@@ -305,6 +309,47 @@ Returns a `Set<string>` of supported file extensions.
...
+import {FileTypeParser} from 'file-type';
+
+const customDetectors = [
+	async tokenizer => {
+		const unicornHeader = [85, 78, 73, 67, 79, 82, 78]; // "UNICORN" as decimal string
+		const buffer = Buffer.alloc(7);
+		await tokenizer.peekBuffer(buffer, {length: unicornHeader.length, mayBeLess: true});
+		if (unicornHeader.every((value, index) => value === buffer[index])) {
+			return {ext: 'unicorn', mime: 'application/unicorn'};
+		}
```
sindresorhus, line 341:
```suggestion

		if (unicornHeader.every((value, index) => value === buffer[index])) {
			return {ext: 'unicorn', mime: 'application/unicorn'};
		}
```

readme.md
```diff
@@ -305,6 +309,47 @@ Returns a `Set<string>` of supported file extensions.
...
+
+If the detector returns `undefined`, there are 2 possible scenarios:
+
+1. The detector has not read from the tokenizer, it will be proceeded with the next available detector.
+2. The detector has read from the tokenizer (`tokenizer.position` has been increased).
+		In that case no further detectors will be executed and the final conclusion is that file-type returns undefined.
+		Note that this an exceptional scenario, as the detector takes the opportunity from any other detector to determine the file type.
+
+
+Example detector array which can be extended and provided to each public method via the fileTypeOptions argument:
```
sindresorhus, line 330:
```suggestion
Example detector array which can be extended and provided to each public method via the `fileTypeOptions` argument:

```

core.d.ts
```diff
@@ -435,3 +438,88 @@ console.log(await fileTypeFromBlob(blob));
 ```
 */
 export declare function fileTypeFromBlob(blob: Blob): Promise<FileTypeResult | undefined>;
+
+/**
+Function that allows specifying custom detection mechanisms.
+
+An iterable of detectors can be provided via the fileTypeOptions argument for the {@link FileTypeParser.constructor}.
```
sindresorhus, line 445:
```suggestion
An iterable of detectors can be provided via the `fileTypeOptions` argument for the {@link FileTypeParser.constructor}.
```

core.d.ts
```diff
@@ -435,3 +438,88 @@ console.log(await fileTypeFromBlob(blob));
...
+
+The detectors are called before the default detections in the provided order.
+
+Custom detectors can be used to add new FileTypeResults or to modify return behaviour of existing FileTypeResult detections.
```
sindresorhus, line 449:
```suggestion
Custom detectors can be used to add new `FileTypeResults` or to modify return behaviour of existing `FileTypeResult` detections.
```

test.js
```diff
@@ -661,3 +663,167 @@ test('corrupt MKV throws', async t => {
...
+test('fileTypeFromTokenizer should return undefined when a custom detector changes the tokenizer position and does not return a file type', async t => {
+	const header = 'UNICORN FILE\n';
+	const uint8ArrayContent = new TextEncoder().encode(header);
+
+	// Include the unicormDetector here to verify it's not used after the tokenizer.position changed
+	const customDetectors = [tokenizerPositionChanger, unicornDetector];
+	const parser = new FileTypeParser({customDetectors});
+
+	const result = await parser.fromTokenizer(strtok3.fromBuffer(uint8ArrayContent));
+	t.true(result === undefined);
```
sindresorhus, line 828:
```suggestion
	is(result, undefined);
```

sindresorhus (changes requested):

core.d.ts
midmarch: I wanted explictly mention the fact that this is the way to parse [File](https://developer.mozilla.org/en-US/docs/Web/API/File), as users keep being surprised the Node.js specific function `fileTypeFromFile` cannot be used in a browser environment.

core.d.ts
Borewit: I wanted explictly mention the fact that this is the way to parse [File](https://developer.mozilla.org/en-US/docs/Web/API/File), as users keep being surprised the Node.js specific function `fileTypeFromFile` cannot be used in a browser environment.

Issue #612, opened 2023-10-26
Password-protected `xlsx` files detected as `cfb`


vecerek:
### Steps to reproduce

TODO

### Expected result

The file is correctly detected as `xlsx`.

### Actual result

The file is incorrectly detected as `cfb`.

Borewit:
[MS-CFB: Microsoft Compound File Binary File Format based format](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cfb/53989ce4-7b05-4f8d-829b-d08d6148375b) are not supported

PR #613, opened 2023-10-26
Add test case for misidentified password-protected xlsx file

vecerek:
This PR serves as a reproduction case for #612.


sindresorhus:
Can you put it in `failingFixture` instead?

Borewit:
I think the identification of file-type is correct, the file provided is a [Compound File Binary Format](https://en.wikipedia.org/wiki/Compound_File_Binary_Format). It maybe a `.xls` format, but certainly not a `.xlsx` which is an [OOXML format](https://en.wikipedia.org/wiki/Office_Open_XML) using ZIP.

PR #615, opened 2023-10-31
Add support for Mach-O

JacobLinCool:
This PR introduces the detection for [Mach-O](https://en.wikipedia.org/wiki/Mach-O), which has been widely used by [Mach](https://en.wikipedia.org/wiki/Mach_kernel)-based operating systems (macOS, iOS, etc.)


Borewit (approved):
LGTM

PR #620, opened 2023-12-30
Use mimetype `audio/wav` for .wav files

luckv:
Use mimetype `audio/wav` for `.wav` files instead of unsupported `audio/vnd.wave`

As pointed out in this [comment](https://github.com/sindresorhus/file-type/pull/150#issuecomment-1869471110), the correct mimetype supported by all major browsers (chrome, firefox, safari) is `audio/wav` and not `audio/vnd.wave`.

I made a simple html page to test the behaviour, which also include a test for `.avi` files. The test show that I didn't find any supported mimetype for `.avi` files, so I will not do a PR for that.

[test-mimetype.zip](https://github.com/sindresorhus/file-type/files/13798200/test-mimetype.zip)

### DIFF
--- a/core.d.ts
+++ b/core.d.ts
@@ -192,7 +192,7 @@ export type MimeType =
 	| 'video/webm'
 	| 'video/quicktime'
 	| 'video/vnd.avi'
-	| 'audio/vnd.wave'
+	| 'audio/wav'
 	| 'audio/qcelp'
 	| 'audio/x-ms-asf'
 	| 'video/x-ms-asf'
--- a/core.js
+++ b/core.js
@@ -835,7 +835,7 @@ export class FileTypeParser {
 			if (this.check([0x57, 0x41, 0x56, 0x45], {offset: 8})) {
 				return {
 					ext: 'wav',
-					mime: 'audio/vnd.wave',
+					mime: 'audio/wav',
 				};
 			}
 
--- a/supported.js
+++ b/supported.js
@@ -189,7 +189,7 @@ export const mimeTypes = [
 	'video/webm',
 	'video/quicktime',
 	'video/vnd.avi',
-	'audio/vnd.wave',
+	'audio/wav',
 	'audio/qcelp',
 	'audio/x-ms-asf',
 	'video/x-ms-asf',


sindresorhus (approved):

Borewit:
Wiki mentions 3 used MIME types for the "WAV audio file format":
1. `audio/vnd.wave`
2. `audio/wav`
3. `audio/wave`
4. `audio/x-wav `

None of those is solid, as there is no authoritative [IANA](https://www.iana.org/) registration.

`audio/vnd.wave` is mentioned in [RFC 2361 WAVE and AVI Codec Registries](https://datatracker.ietf.org/doc/html/rfc2361)

`vnd.` is prefixed for _vendor_ as specified in [RFC 6838 Media Type Specifications and Registration Procedures](https://www.rfc-editor.org/rfc/rfc6838.html#page-6). After the `.vnd` the vendor name should appear, so `audio/vnd.microsoft.wave` would be a more sensible MIME type in my humble opinion. Given that the WAV format is in nowadays more common good then proprietary. 

WAV is using the [RIFF container format](https://en.wikipedia.org/wiki/Resource_Interchange_File_Format), which is documented with a unregistered MIME type: `application/x-riff`. To fall back on the container format, does not seem to be the convention.

Unregistered entries should be prefixed `x-` (no 4) [*](https://www.rfc-editor.org/rfc/rfc6838.html#section-3.4). By that definition, number four is valid. But the The paradoxical statement made also in [RFC 6838](https://www.rfc-editor.org/rfc/rfc6838.html#section-3.4)  is that these `x-` should only be used in `private` / `local` environments. The reality is, is that WAV files are widely used in public, and just to prevent this kind of painful solution, if I would be working for IANA, I would just register `audio/wav` for the sake of normalization. Yet I do not work for IANA.

Then there is `audio/wav`. `wav` is the original Microsoft extension, in the MIME types there is no need to stick to the 3 character extension, and in that aspect `audio/wave` or `audio/waveform` would be more logic. 

I have no objections to change to `audio/wav`, it's a bit of an arbitrary choice. One can choose the most widely spread convention, unfortunately this is also something which is hard to measure.

If the [Mozilla listing of `audio/wav`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) can be accepted as a common browser convention, we could follow their convention as well.

Borewit (approved):

Borewit:
Sorry, I noticed the discussion on #150 after I wrote my response. Conclusion is aligned.

Issue #578, opened 2023-02-06
 Consider alternate (browser friendly) buffer


rolanday:
Hello! :-) I'd love to be able to use this module in my SvelteKit app, but node:buffer dep is not something I've been able to workaround via Vite polyfill (same issue for other frameworks like Nextjs as well). 

Have or can you consider alt buffer impl that is browser friendly? Or, any tips on configuring for use in SvelteKit, Vue, Nextjs, etc?

Thanks!

sindresorhus:
Ideally, all code should be using `Uint8Array` instead of `Buffer`, but there's just too much Node.js code that depends on `Buffer`.

If anyone wants to work on this, the first task would be to convert the dependencies of `file-type` to use `Uint8Array` instead of `Buffer`.

volarname:
+1 
this library really needs a browser version without node and webpack deps
something like `fromFile(file: File)`
because file.mimeType doesnt seems to work properly all times

Borewit:
> +1 this library really needs a browser version without node and webpack deps something like `fromFile(file: File)` because file.mimeType doesnt seems to work properly all times

That function already exist @volarname :  [`fileTypeFromBlob(blob`)](https://github.com/sindresorhus/file-type#filetypefromblobblob). Where you can pass you file as blob, since [File](https://developer.mozilla.org/en-US/docs/Web/API/File) inherits from [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob).

Borewit:
This module has been primary written for Node.js. It does depend on Buffer. There is [browser friendly buffer](https://www.npmjs.com/package/buffer) module available, but you depend on your framework how easy that is to fit it in. 
 There is not much more we can do, as these core dependencies are not shared between browser and Node.js.

sindresorhus:
https://sindresorhus.com/blog/goodbye-nodejs-buffer

Borewit:
> https://sindresorhus.com/blog/goodbye-nodejs-buffer

Interesting!, very nice guidance. I will give it try (starting with the dependent modules first).

Borewit:
It is not only `node:buffer` `file-type` depends on, also `safe-buffer`
```mermaid
graph TD;
    FT-->S
    FT(file-type)-->RWNS(readable-web-to-node-stream)
    S(strtok3)-->P(peek-readable)
    S(strtok3)-->TO("@tokenizer/token")
    TY(token-types)-->TO
    TY-->IE("ieee754")
[...]

PR #633, opened 2024-06-27
Replace node:buffer usage with Uint8Array

bjornstar:
It's time to say [Goodbye, node.js Buffer](https://sindresorhus.com/blog/goodbye-nodejs-buffer) in this library.

Resolves #578

Dependencies
=================

- [x] [Update `token-types` to take encoding for `StringType`](https://github.com/Borewit/token-types/pull/650)
- [x] [Add `indexOf` & `getUintBE` to `uint8array-extras`](https://github.com/sindresorhus/uint8array-extras/pull/12)

~This was an interesting challenge. Please hold while I raise PRs to these two dependencies.~

~Update: PRs raised for both dependencies, I'll come back and update this once they're merged and released.~

Update: Both dependencies have been updated and released.

PR #635, opened 2024-07-04
Add support for reading from a WebStreams

### LINKED ISSUE #594: Make `Blob` lazy
See: https://github.com/sindresorhus/file-type/pull/588#issuecomment-1493880124

### LINKED ISSUE #617: Support reading from browser streams in server-side export
In SvelteKit, the `event.request.body` is a web-interoperable stream. However, when this package is imported in a SvelteKit project running in Bun, it tries to load the NodeJS stuff.

I recommend moving the NodeJS-related legacy stuff into a `file-type/node` package, as the majority is moving towards modern standards with runtimes like Bun and Deno which support WinterCG stuff, and SSR frameworks like SvelteKit also support it.

Borewit:
Changes:

- Add support for reading from a WebStreams
- Stream Blob via a WebStreams, instead of buffering the full content

Moves remaining native Node Readable streams to the Node.js specific entry point (`index.js`). As the fileTypeStream is based on the the native Node Readable stream, that function is no longer available in the primary (other then Node engine) entry point (`core.js`).

:warning: `fromBlob()` requires Node.js ≥ 20 as the `strtok3.fromStream` requires a Readable from which a [ReadableStreamBYOBReader](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamBYOBReader) can be created. Which is only available since Node.js 20.

- Resolves #594 
- Resolves #617
- Depends on: https://github.com/Borewit/strtok3/pull/1098

#### Updating dependencies to:

```mermaid
graph TD;
    FTN("file-type (Node.js entry point)")-->FTP
    FTP("file-type (primary entry point)")-->S
    S(strtok3)-->P(peek-readable)
    S(strtok3)-->TO("@tokenizer/token")
    TY(token-types)-->TO
    TY-->IE("ieee754")
    FTP-->TY
    NS("node:stream")
    FTN-->NS
    FTP-->UAE(uint8array-extras)
    style NS fill:#F88,stroke:#A44
    style IE fill:#CCC,stroke:#888
    style FTN fill:#FAA,stroke:#A44
```
Previous dependency diagram can be found here: https://github.com/sindresorhus/file-type/issues/578#issuecomment-1789138382


sindresorhus:
I would prefer to have a single `fromStream` method that supports both.

Borewit:
> I would prefer to have a single `fromStream` method that supports both.

That was quick.
I will give that some thought, I like to release this one together #633, as joined effort to work without Node Streams and Buffer.

Borewit:
> I would prefer to have a single `fromStream` method that supports both.

@sindresorhus, are you not concerned that this make it more difficult to for tree-shaking to exclude `node:buffer` and `nodes:stream`  when these are effectively not being used?

sindresorhus:
I was thinking that `fromStream` in `core.js` would only support web streams, same with the one in `browser.js`, but in `index.js` it would get the additional support for Node.js streams. That way, only Node.js users gets the Node.js imports.

sindresorhus:
Alternatively, we could just drop support for Node.js streams, and just tell users to call `.toWeb()` on their Node.js streams. That may be even better actually.

sindresorhus:
> Alternatively, we could just drop support for Node.js streams, and just tell users to call .toWeb() on their Node.js streams. That may be even better actually.

Let's go with this.

Borewit:
> > Alternatively, we could just drop support for Node.js streams, and just tell users to call .toWeb() on their Node.js streams. That may be even better actually.
> 
> Let's go with this.

The [`toWeb()` method] (https://nodejs.org/api/stream.html#streamreadabletowebstreamreadable) is still experimental.

Maybe first release with as a hybrid solution, including Node.js & Web Streams.  And indeed, moving the Node.js dependencies to the Node.js specific entry point `index.js` . Also note that in addition to the `fileTypeFromStream()`, there is the `fileStream()` which is still depends on Node.js streams, and also shared to with the primary entry point. I am honestly not very motivated to update that method, as it not complaint with the current architecture where file-type has access to the entire file. The argument for backward compatibility is now fading, as Web Streams are now becoming widely available.

sindresorhus:
> Maybe first release with as a hybrid solution, including Node.js & Web Streams. And indeed, moving the Node.js dependencies to the Node.js specific entry point index.js .

Ok, sure. But at least add a `TODO` comment to remove Node.js streams support and recommend `toWeb()` instead when it's stable.

sindresorhus:
> there is the fileStream() which is still depends on Node.js streams

It could be moved to `index.js` as it's already documented to only be available on Node.js.

core.d.ts
```diff
@@ -1,4 +1,9 @@
-import type {Readable as ReadableStream} from 'node:stream';
+/**
+ * Typings for primary entry point, Node.js specific typings can be found in index.d.ts
+ */
+
+import type {ReadableStream as WebReadableStream} from 'node:stream/web';
```
sindresorhus, line 5: This should be available globally.

core.d.ts
```diff
@@ -511,7 +516,7 @@ export declare class FileTypeParser {
 	/**
 	Works the same way as {@link fileTypeFromStream}, additionally taking into account custom detectors (if any were provided to the constructor).
 	*/
-	fromStream(stream: ReadableStream): Promise<FileTypeResult | undefined>;
+	fromStream(stream: NodeReadableStream): Promise<FileTypeResult | undefined>;
```
sindresorhus, line 519: Shouldn't this and `toDetectionStream` support web streams too?

sindresorhus (changes requested):

core.d.ts
Borewit: The global type in incompatible with Node.js type. If I change it here, the problem will appear elsewhere,

core.d.ts
Borewit: Yeah, should probably be done as well.

core.d.ts
sindresorhus: I'm not sure what that means, but this is `core.d.ts`, so it shouldn't import types only available for Node.js.

Borewit: The Node.js Readable types are incompatible with the global `lib.dom.d.ts` types, one things the challenges mentioned in https://github.com/sindresorhus/file-type/pull/588#issuecomment-1493880124, related to this issue https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60377, which unfortunately get closed with PR aiming to resolve that.

I used the Node.js, as Node.js has been the primary supported platform. Using the `lib.dom.d.ts` I need hack in types mapping cast somewhere.

Maybe good accept both, an do an ugly type cast, just for the convenience for our users.

Off-topic: The other mind f*ck, was the [BYOB Readable Stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamBYOBReader), The "Bring-Your-Own-Buffer" reader:
> defines a reader for a [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) that supports **zero-copy** reading from an underlying byte source

Well the first things the **zero-copy** method does is hijacking _the buffer_ you bring, and essentially turns into junk (there is formal property for this state, forgot the name), meaning it can no longer be used. Which essentially forces you to create a new Buffer (as it becomes totally useless after providing it), and then copying the data to the Buffer you wanted to have it written in the first place. So the only feature BYOB has, versus the [ReadableStreamDefaultReader](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader) is that you can control the chunk length to be written. How confusing, what a disappointment.

Borewit:
> > there is the fileStream() which is still depends on Node.js streams
> 
> It could be moved to `index.js` as it's already documented to only be available on Node.js.

That is very doable. We can raise an issue, to implement the same concept using Web Streams. I actually starting working on that, but it takes some time. Tricky stuff those streams.

sindresorhus:
@Borewit Are we ready to do another release now or do you plan more changes? None of these changes are technically breaking, right? Since Buffer is also a Uint8Array, so https://github.com/sindresorhus/file-type/commit/00e051bceaf0791ffc1b08b36aee196ccdd95606 should not be breaking (it's only breaking if we would have changed a return value from Buffer to Uint8Array).

Borewit:
> @Borewit Are we ready to do another release now or do you plan more changes? None of these changes are technically breaking, right? Since Buffer is also a Uint8Array, so [00e051b](https://github.com/sindresorhus/file-type/commit/00e051bceaf0791ffc1b08b36aee196ccdd95606) should not be breaking (it's only breaking if we would have changed a return value from Buffer to Uint8Array).

This morning I realized, what could be documented better is the `NodeFileTypeParser` class

https://github.com/sindresorhus/file-type/blob/b815b5e173dd9f521311615eb33e955181030ca9/index.js#L9

which should be used instead of the `FileTypeParser`, for those implementing custom extensions and usign Node.js.

No objection to release now.

Borewit:
Part of [v19.1.0](https://github.com/sindresorhus/file-type/releases/tag/v19.1.0)

PR #644, opened 2024-07-11
Improve class method consistency, adding `NodeFileTypeParser.fromFile()`

Borewit:
Changes:
- Improve class method consistency, adding `NodeFileTypeParser.fromFile()`
- Add missing parameter in `fileTypeFromFile`

### DIFF
--- a/core.d.ts
+++ b/core.d.ts
@@ -476,6 +476,9 @@ export declare class TokenizerPositionError extends Error {
 export declare class FileTypeParser {
 	detectors: Iterable<Detector>;
 
+	/**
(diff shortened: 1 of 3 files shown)


core.d.ts
```diff
@@ -476,6 +476,9 @@ export declare class TokenizerPositionError extends Error {
 export declare class FileTypeParser {
 	detectors: Iterable<Detector>;
 
+	/**
+	 @param options -  `options.customDetectors`: list of custom detectors
+	 */
```
sindresorhus, line 481: Doc comments are not needed if they don't add anything to the types. Here, the doc comments just repeats the information available in the below types.

PR #649, opened 2024-07-29
Add detecting Web Stream to default (core) entry point

Borewit:
In addition to the `fileTypeStream()` converting a Node Readable stream, into a detecting stream, add support for web streams.

In Node.js you can either use a Node Readable stream, or a Web Stream.
Using the default entry point (core), just supports Web Stream.

### DIFF
--- a/core.js
+++ b/core.js
@@ -51,7 +51,7 @@ export async function fileTypeFromTokenizer(tokenizer) {
 	return new FileTypeParser().fromTokenizer(tokenizer);
 }
 
-export async function stream(webStream) {
+export async function fileTypeStream(webStream) {
 	return new FileTypeParser().toDetectionStream(webStream);
 }
 
--- a/core.js
+++ b/core.js
@@ -51,6 +51,10 @@ export async function fileTypeFromTokenizer(tokenizer) {
 	return new FileTypeParser().fromTokenizer(tokenizer);
 }
 
+export async function stream(webStream) {
+	return new FileTypeParser().toDetectionStream(webStream);
+}
+
 export class FileTypeParser {
 	constructor(options) {
 		this.detectors = options?.customDetectors;
@@ -102,6 +106,50 @@ export class FileTypeParser {
 		} finally {
 			await tokenizer.close();
 		}
+	}
+
+	async toDetectionStream(webStream, options = {}) {
+		const {sampleSize = reasonableDetectionSizeInBytes} = options;
+
+		// Initialize a reader from the web stream
+		const reader = webStream.getReader({mode: 'byob'});
+		const pass = new TransformStream();
+		const writer = pass.writable.getWriter();
+		let detectedFileType;
+
+		// Read the first chunk for file type detection
+		const {value: chunk, done} = await reader.read(new Uint8Array(sampleSize));
+		if (done || !chunk) {
+			detectedFileType = undefined;
+		} else {
+			try {
+				detectedFileType = await this.fromBuffer(chunk.slice(0, sampleSize));
+			} catch (error) {
+				if (error instanceof strtok3.EndOfStreamError) {
+					detectedFileType = undefined;
+				} else {
+					throw error;
+				}
+			}
+		}
+
+		// Write the initial chunk into the pass-through stream
+		writer.write(chunk);
+
+		// Forward remaining data from the reader to the writer
+		(async function pump() {
+			const {value, done} = await reader.read(new Uint8Array(512 * 1024));
+			if (done) {
+				return writer.close();
+			}
+
+			await writer.write(value);
+			return pump();
+		})();
+
+		// Attach the detected file type to the output stream
+		pass.readable.fileType = detectedFileType;
+		return pass.readable;
 	}
 
 	check(header, options) {
--- a/core.js
+++ b/core.js
@@ -138,6 +138,7 @@ export class FileTypeParser {
 
 		// Forward remaining data from the reader to the writer
 		(async function pump() {
+			// Read a 512 kB chunk, where the chunk size is an instinctively chosen value
 			const {value, done} = await reader.read(new Uint8Array(512 * 1024));
 			if (done) {
 				return writer.close();
--- a/core.js
+++ b/core.js
@@ -108,57 +108,49 @@ export class FileTypeParser {
 		}
 	}
 
-	async toDetectionStream(webStream, options = {}) {
+	async toDetectionStream(stream, options) {
 		const {sampleSize = reasonableDetectionSizeInBytes} = options;
-
-		// Initialize a reader from the web stream
-		const reader = webStream.getReader({mode: 'byob'});
-		const pass = new TransformStream();
-		const writer = pass.writable.getWriter();
 		let detectedFileType;
+		let firstChunk;
 
-		// Read the first chunk for file type detection
-		const {value: chunk, done} = await reader.read(new Uint8Array(sampleSize));
-		if (done || !chunk) {
-			detectedFileType = undefined;
-		} else {
-			try {
-				detectedFileType = await this.fromBuffer(chunk.slice(0, sampleSize));
-			} catch (error) {
-				if (error instanceof strtok3.EndOfStreamError) {
+		const reader = stream.getReader({mode: 'byob'});
+		try {
+			// Read the first chunk from the stream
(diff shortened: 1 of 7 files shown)


Borewit:
Please merge #650 & #651 first.

Borewit:
This PR undo PR #651, as the constructor is not overridden, there is no need to re-declare it.

index.d.ts
```diff
@@ -70,10 +71,10 @@ A smaller sample size will result in lower probability of the best file type det
 **Note:** Requires Node.js 14 or later.
 
 @param readableStream - A [readable stream](https://nodejs.org/api/stream.html#stream_class_stream_readable) containing a file to examine.
-@param options - Maybe used to override the default sample-size.
-@returns A `Promise` which resolves to the original readable stream argument, but with an added `fileType` property, which is an object like the one returned from `fileTypeFromFile()`.
+@param options - Maybe used to override the default sample-size.with an added `fileType` property, which is an object like the one returned from `fileTypeFromFile()`.
 
 @example
+@returns A `Promise` which resolves to the original readable stream argument, but
```
sindresorhus, line 77: Incomplete sentence. And in the wrong place.

index.js
```diff
@@ -36,8 +40,8 @@ export class NodeFileTypeParser extends FileTypeParser {
 				(async () => {
 					try {
 						// Set up output stream
-						const pass = new stream.PassThrough();
-						const outputStream = stream.pipeline ? stream.pipeline(readableStream, pass, () => {}) : readableStream.pipe(pass);
+						const pass = new PassThrough();
+						const outputStream = pipeline ? pipeline(readableStream, pass, () => {}) : readableStream.pipe(pass);
```
sindresorhus, line 44: `() => {}` means you are effectively ignoring any pipeline errors.

test.js
```diff
@@ -337,7 +338,7 @@ async function testFileNodeFromStream(t, ext, name) {
 	t.is(typeof fileType.mime, 'string', 'fileType.mime');
 }
 
-async function loadEntireFile(readable) {
+async function loadEntireFileFromNodeReadable(readable) {
```
sindresorhus, line 341: You can just use https://github.com/sindresorhus/get-stream for this and `loadEntireFileFromWebStream`. No need for custom logic.

core.js
```diff
@@ -51,6 +51,10 @@ export async function fileTypeFromTokenizer(tokenizer) {
 	return new FileTypeParser().fromTokenizer(tokenizer);
 }
 
+export async function stream(webStream) {
```
sindresorhus, line 54: Wrong name.

core.js
```diff
@@ -104,6 +108,50 @@ export class FileTypeParser {
...
+
+		// Forward remaining data from the reader to the writer
+		(async function pump() {
+			const {value, done} = await reader.read(new Uint8Array(512 * 1024));
+			if (done) {
+				return writer.close();
+			}
+
+			await writer.write(value);
+			return pump();
```
sindresorhus, line 147: It will cause a unhandled rejection if anything in here throws. Should be properly handled. Also, returning the values are moot as the return values are not used.

core.js
```diff
@@ -104,6 +108,50 @@ export class FileTypeParser {
...
+				}
+			}
+		}
+
+		// Write the initial chunk into the pass-through stream
+		writer.write(chunk);
+
+		// Forward remaining data from the reader to the writer
+		(async function pump() {
+			const {value, done} = await reader.read(new Uint8Array(512 * 1024));
```
sindresorhus, line 141: The magic number should have a code comment explaining it.

sindresorhus (changes requested):

sindresorhus:
I have committed some changes, so please don't force push and override them.

sindresorhus:
Ensure the readme and index.d.ts are in sync.

index.js
Borewit: Not my code

core.js
```diff
@@ -104,6 +108,59 @@ export class FileTypeParser {
...
+				writer.abort(readError); // Abort writing on error
+			});
+			if (done) {
+				return writer.close();
+			}
+
+			await writer.write(value).catch(writeError => {
+				reader.cancel(writeError); // Cancel the reader on error
+			});
+			return pump(); // Recursion for continuous reading
```
sindresorhus, line 156:
```suggestion
			await pump(); // Recursion for continuous reading
```

core.js
```diff
@@ -104,6 +108,59 @@ export class FileTypeParser {
...
+			const {value, done} = await reader.read(new Uint8Array(512 * 1024)).catch(readError => {
+				writer.abort(readError); // Abort writing on error
+			});
+			if (done) {
+				return writer.close();
+			}
+
+			await writer.write(value).catch(writeError => {
+				reader.cancel(writeError); // Cancel the reader on error
+			});
```
sindresorhus, line 155: Use try/catch, not `.catch` on the promise.

core.js
```diff
@@ -104,6 +108,59 @@ export class FileTypeParser {
 		}
 	}
 
+	async toDetectionStream(webStream, options = {}) {
```
sindresorhus, line 111: I think this could be made simpler by using `pipeThrough`.


Example code (untested):


```js
function toDetectionStream(stream) {
	let buffer = new Uint8Array();
	let detectedFileType;

	const transformStream = new TransformStream({
		async start(controller) {
			const reader = stream.getReader();
			let bytesRead = 0;

			const {value: chunk, done} = await reader.read(new Uint8Array(sampleSize));
			if (done || !chunk) {
				detectedFileType = undefined;
			} else {
				try {
					detectedFileType = await this.fromBuffer(chunk.slice(0, sampleSize));
				} catch (error) {
					if (error instanceof strtok3.EndOfStreamError) {
						detectedFileType = undefined;
					} else {
						throw error;
					}
				}
			}

			controller.enqueue(value);
		},
		transform(chunk, controller) {
			controller.enqueue(chunk);
		}
	});

	const finalStream = stream.pipeThrough(transformStream);

	finalStream.fileType = detectedFileType;

	return finalStream;
}
```

sindresorhus (changes requested):

core.js
Borewit: That would be great if that would work.
But when running `const finalStream = stream.pipeThrough(transformStream);`, it throws this error:
```
TypeError [ERR_INVALID_STATE]: Invalid state: The ReadableStream is locked
```
I tried to release the reader with `reader.releaseLock();`, in `start(controller)`, before the `pipeThrough`, but that did not resolve the issue.

See branch [try-determine-file-type-using-transform-stream](https://github.com/sindresorhus/file-type/tree/try-determine-file-type-using-transform-stream), commit 1bb56850263610585ccf0b13144d10f78fd78993.

Borewit: Got it working, by taking out the file-type detection of the transformation start(), to ensure the stream is unlocked before starting the pipeThrough(): 53ec046d56f496f41eed5f37f8b1faffaea2b428

index.d.ts
```diff
@@ -61,19 +60,19 @@ export function fileTypeFromStream(stream: AnyWebReadableStream<Uint8Array> | No
...
 
-@param readableStream - A [readable stream](https://nodejs.org/api/stream.html#stream_class_stream_readable) containing a file to examine.
-@param options - Maybe used to override the default sample-size.
-@returns A `Promise` which resolves to the original readable stream argument, but with an added `fileType` property, which is an object like the one returned from `fileTypeFromFile()`.
-
-@example
+@param readableStream - A [Node.js `stream.Readable`](https://nodejs.org/api/stream.html#stream_class_stream_readable)
+or [web `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream), streaming a file to examine.
+@param options - May be used to override the default sample size with an added `fileType` property, wh` property, which is an object like the one returned from `fileTypeFromFile()`.
+@exampleich is an object like the one returned from `fileTypeFromFile()`.
```
sindresorhus, line 74: Some typos here.

index.d.ts
```diff
@@ -61,19 +60,19 @@ export function fileTypeFromStream(stream: AnyWebReadableStream<Uint8Array> | No
...
+@returns A `Promise` which resolves to the original readable stream argument, but with an added `fileType
```
sindresorhus, line 75: This should be above `@example`.

sindresorhus: And it's incomplete.

Issue #657, opened 2024-08-14
Support for WebVTT files (text/vtt)


AleksandrHovhannisyan:
First, thanks so much for this package! We've been using it at work to validate files uploaded by users and it works as expected for the majority of our use cases. There is one edge case where it doesn't currently validate WebVTT files (MIME type `text/vtt`, for captions shown in a video element's `<track>`).

The magic numbers for VTT files are as follows according to the W3 document titled [WebVTT: The Web Video Text Tracks Format](https://www.w3.org/TR/webvtt1/#iana-text-vtt):

> WebVTT files all begin with one of the following byte sequences (where "EOF" means the end of the file):
>
> EF BB BF 57 45 42 56 54 54 0A
> EF BB BF 57 45 42 56 54 54 0D
> EF BB BF 57 45 42 56 54 54 20
> EF BB BF 57 45 42 56 54 54 09
> EF BB BF 57 45 42 56 54 54 EOF
> 57 45 42 56 54 54 0A
> 57 45 42 56 54 54 0D
> 57 45 42 56 54 54 20
> 57 45 42 56 54 54 09
> 57 45 42 56 54 54 EOF
> (An optional UTF-8 BOM, the ASCII string "WEBVTT", and finally a space, tab, line break, or the end of the file.)

Would it be possible to support this? If so, I'd be happy to help or put in a PR.

Borewit:
That is in my opinion in scope.

Please note that we got BOM covered in a generic way:

https://github.com/sindresorhus/file-type/blob/988bf4bc9f9bc98e8f3360da4dfa36e5caa455b3/core.js#L251-L255

So ignore the magic numbers with the BOM field (`EF BB BF`), those will be automatically covered.

I suggest to trigger on `WEBVTT`, and possibly match the last character.

AleksandrHovhannisyan:
Thanks! That makes sense. I'll work on this and put up a PR.

PR #658, opened 2024-08-14
Add support for WebVTT

AleksandrHovhannisyan:
Resolves: https://github.com/sindresorhus/file-type/issues/657

Docs/magic numbers: https://www.w3.org/TR/webvtt1/#iana-text-vtt


core.js
```diff
@@ -1139,6 +1139,19 @@ export class FileTypeParser {
...
+			|| this.check([0x57, 0x45, 0x42, 0x56, 0x54, 0x54, 0x0A])	// LF
+			|| this.check([0x57, 0x45, 0x42, 0x56, 0x54, 0x54, 0x0D])	// CR
+			|| this.check([0x57, 0x45, 0x42, 0x56, 0x54, 0x54, 0x09])	// Tab
+			|| this.check([0x57, 0x45, 0x42, 0x56, 0x54, 0x54, 0x20])	// Space
+		) {
+			return {
+				ext: 'vtt',
+				mime: 'text/vtt',
+			};
+		}
```
AleksandrHovhannisyan, line 1153: Is this a sensible way of going about it, or should I be testing the common byte sequence first and then the last byte in a nested statement? Also, does it make sense to have separate fixtures for these different characters after the initial `WEBVTT`?

AleksandrHovhannisyan: That makes sense, thanks! I'll make these changes.

Btw, is there any code formatter set up for this project? I couldn't find anything in the docs or package.json. Fixing the tests is a bit frustrating as it's getting hung up on code formatting.

sindresorhus:
```
npx xo --fix
```

core.js
```diff
@@ -1140,11 +1140,14 @@ export class FileTypeParser {
...
-			|| this.check([0x57, 0x45, 0x42, 0x56, 0x54, 0x54, 0x09])	// WEBVTT<Tab>
-			|| this.check([0x57, 0x45, 0x42, 0x56, 0x54, 0x54, 0x20])	// WEBVTT<Space>
+			this.checkString('WEBVTT')
+			&& (
+				this.check([0x00], {offset: 6})		// EOF
+				|| this.check([0x0A], {offset: 6})	// LF
+				|| this.check([0x0D], {offset: 6})	// CR
+				|| this.check([0x09], {offset: 6})	// Tab
+				|| this.check([0x20], {offset: 6})	// Space
+			)
```
AleksandrHovhannisyan, line 1150: Sorry that I'm asking so many questions: Should I be checking some of these as strings/escape sequences too since they're all ASCII? Also, is there a way to avoid repeating the offset? I tried doing an `await tokenizer.ignore(6)` after the `WEBVTT` check but that didn't seem to work.

Borewit: If you call `await tokenizer.ignore(6)`, you permanently move the offset with 6 bytes. That also require you need to re-read the buffer used by `this.check`, starting on that offset.

But, that approach is irreversible, meaning all following tests will fail, as they can not ,longer read from offset 0.
Therefor, if you read of (or ignore) from the tokenizer, you have to return the file type or return undefined. You essentially can only do that, if you are sure that remaining tests are pointless to run.

Your code looks fine to me, maybe a bit lengthy as a result of the repeating offset argument, but very readable.

If you prefere you could shorten it by putting the special character in an array, and calling [every] (not tested):

```js
[0x00, 0x0A, 0x0D, 0x09, 0x20].every(lastChar => this.check([lastChar], {offset: 6}))
```

> Should I be checking some of these as strings/escape sequences

I think this is fine without, but you can try, and see if gets more elegant.

AleksandrHovhannisyan: Thanks for the help!

Borewit (commented):
The 2 tests before your tests are actually not testing within the 7 first bytes. But that is different issue.

readme.md
```diff
@@ -507,6 +507,7 @@ console.log(fileType);
 - [`vcf`](https://en.wikipedia.org/wiki/VCard) - vCard
 - [`voc`](https://wiki.multimedia.cx/index.php/Creative_Voice) - Creative Voice File
 - [`vsdx`](https://en.wikipedia.org/wiki/Microsoft_Visio) - Microsoft Visio File
+- [`vtt`](https://www.w3.org/TR/webvtt1/) - WebVTT File (for video captions)
```
Borewit, line 510: It looks like the convention is to link to wikipedia 
```suggestion
- [`vtt`](https://en.wikipedia.org/wiki/WebVTT) - Web Video Text Tracks, a W3C standard file for subtitles or captures
```

Borewit (changes requested):

readme.md
AleksandrHovhannisyan: Yup, I noticed that too. The only reason I didn't do that is because the Wikipedia article doesn't list the magic numbers for this file format, so I thought it would be more helpful to link to the w3 spec.

fixture/fixture-vtt-eof.vtt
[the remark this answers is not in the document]
AleksandrHovhannisyan: I see—although that makes me wonder: why does the test pass for the EOF fixture? Maybe I'm misunderstanding how the null character behaves in `this.check`.

Is there a method that would tell me if the tokenizer has reached the end of the file? I see it has a `position` property as well as some `peekX` methods.

fixture/fixture-vtt-space.vtt
```diff
@@ -0,0 +1,38 @@
+WEBVTT 00:11.000 --> 00:13.000
+<v Roger Bingham>We are in New York City
```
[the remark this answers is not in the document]
AleksandrHovhannisyan: I didn't test the file in an actual `<track>` (don't think I need to) but I just grabbed it from the W3 doc. See [A simple caption file](https://www.w3.org/TR/webvtt1/#introduction-caption)).

readme.md
Borewit: It also the formal standard, so I fully get you. But it is not lost, it remains documented as part of this PR.

I think the function of this link, in addition as a reference, the links servers to give an context to ensure there can be no ambiguity which file type we are talking about.

fixture/fixture-vtt-eof.vtt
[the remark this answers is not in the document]
Borewit: I was puzzled but that one. Chunks of the file are read, from smaller to larger (that is why we want to have texted ordered by fixture, starting count from 0 offset). 

If a file is read, smaller then the provided buffer size (EOF is ignored with `mayBeLess`).

https://github.com/sindresorhus/file-type/blob/988bf4bc9f9bc98e8f3360da4dfa36e5caa455b3/core.js#L174

Remaining buffer is set to zero, hence you zero, hence your test match.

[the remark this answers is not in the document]
AleksandrHovhannisyan: Gotcha. What would you recommend we do here? I want to make sure I still test for the EOF edge case, even though it's admittedly rare in practice (nobody should upload an empty caption file). That said, I don't think it would be enough to test for just `WEBVTT` as a file that starts with `WEBVTT-not-webvtt`, for example, would pass the test but doesn't comply with the spec. Is there a way to do an exact-length check?

Borewit:
@AleksandrHovhannisyan, I propose to keep the single fixture as you got it from WRC and drop the reverse engineered fixtures to aiming to match each signature cases. Keep it simple.

AleksandrHovhannisyan:
@Borewit Okay, happy to do that, although I will say I found the fixtures helpful when writing the code, as I wrote those first and ran tests as I made adjustments. Without those fixtures, there's no way to guarantee that the code works for edge cases.

Also, I think it would help both of us if I could get a single set of change requests for this PR. Happy to push whatever changes are needed to get it merged, but I do want to limit back and forth.

fixture/fixture-vtt-eof.vtt
Borewit, line 1: There is no leading null character, and a null character is not the same as a EOF.

Borewit (changes requested):

AleksandrHovhannisyan:
Hmm, not sure why tests are failing for Node 20+. Node 18 works fine locally for me.

What is the procedure for running those same CI tests locally? `npm run test` only runs on my local Node 18.

Borewit:
> What is the procedure for running those same CI tests locally?

You essentially need to install a different version of Node or use some fancy mechanism to switch Node versions.

Some tests are only performed on Node >= 20

AleksandrHovhannisyan:
Okay that's fine, I have nvm/fnm so I'll just use one of those. Thanks

core.js
```diff
@@ -1139,18 +1139,32 @@ export class FileTypeParser {
...
-				mime: 'text/vtt',
-			};
+		if (this.checkString('WEBVTT')) {
+			await tokenizer.ignore(6);
+			let isValidWebVTT = false;
+
+			// Try-catch to handle valid edge case of "WEBVTT<EOF>"
+			try {
+				// The WebVTT standard says that the first line should be "WEBVTT" followed by a single ASCII whitespace character
+				const whitespaceToken = await tokenizer.readToken(new Token.StringType(1, 'ascii'));
```
Borewit, line 1149: It does not work like this, you are reading the very from the stream.
Secondly, if you start consuming the stream, all following on test will fail, so you have to determine the file type.

Borewit (changes requested):

core.js
AleksandrHovhannisyan: > The reason you see the following tests did not break is because this rare situation only occures on files starting with WEBVTT, but that does not make it right.

What if someone creates a file that has `WEBVTTx` as the first seven bytes? The first six bytes aren't the full signature per the spec: "An optional UTF-8 BOM, the ASCII string "WEBVTT", and finally a space, tab, line break, or the end of the file." So I don't think it's rigorous enough to just check the first six bytes as it doesn't tell us anything about the seventh byte. With my current code, a file that starts with `WEBVTT` but is _not_ followed by a whitespace character would return `undefined`, which I think is the expected behavior.

> The 7th character you are interested in has already been peeked, the EOF has already been handled.
I know you lost the info where the EOF is, but both \0 and EOF are acceptance, there is no is issue simply testing for \0 is fine.

Sorry, I must have misunderstood you before. I thought you didn't want me to test for `\0` since it's not the same as an EOF.

Borewit: > What if someone creates a file that has WEBVTTx as the first seven bytes? The first six bytes aren't the full signature per the spec: "An optional UTF-8 BOM, the ASCII string "WEBVTT", and finally a space, tab, line break, or the end of the file." So I don't think it's rigorous enough to just check the first six bytes as it doesn't tell us anything about the seventh byte. With my current code, a file that starts with WEBVTT but is not followed by a whitespace character would return undefined, which I think is the expected behavior (because that file does not comply with the signature format for valid WebVTT files).

That is indeed the best solution, but you cannot do it by reading from stream, as that irreversible consumes the stream. If you do it by reading, you break the case of "The first six bytes aren't the full signature".

> Sorry, I must have misunderstood you before. I thought you didn't want me to test for \0 since it's not the same as an EOF.

I am sorry, as I thought `\0` was defined as a formal separator. But it is not, and you are testing on `\0` as that is the way the EOF translates in `file-type`. That is why complained about the EOF fixture file, I thought it had to to have `\0` separator.

AleksandrHovhannisyan: > `WEBVTT78WEBP123XXXX`. This should be not be recognized as `WebVTT`, but it should be recognized as `WEBP`.

Oh, I see. Does this mean the `file-type` parser ignores unrecognized byte sequences at the start of a file and continues searching until it eventually finds a valid signature? I thought magic numbers had to appear at the start of a file.

> I am sorry, as I thought \0 was defined as a formal separator. But it is not, and you are testing on \0 as that is the way the EOF translates in file-type. That is why complained about the EOF fixture file, I thought it had to to have \0 separator.

All good, I think I understand what you meant now.

Borewit: >  Does this mean the file-type parser ignores unrecognized byte sequences at the start of a file and continues searching until it eventually finds a valid signature?

Exactly.

> Edit: So just to be clear, do you want me to simplify the code to just check WEBVTT, without checking the 7th byte, and remove the other fixtures?

No I agree with you to include the 7th.

This commit was very close to perfection 😉 : 53be4b412b4b32650fb66f926961c39bd896e7ea
- Rename variable `whitespace` to `char7`
- change '\r\n` into `\r`
- Using `tokenizer.fileInfo.size === 6` is tricky, it is possible to let file-size read from a stream with an unknown length. Safer to use the `0`, assuming that is the result of an "EOF". 

, and I think it is good to go.

Borewit (approved):
LGTM

Issue #661, opened 2024-08-22
ERR_PACKAGE_PATH_NOT_EXPORTED when file-type used in nest.js


robertok-weiji:
Hi,

I'm using file-type in nest.js and I've this error, using the last 19.4.1 version.

```
nest start
node:internal/modules/cjs/loader:597
      throw e;
      ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in D:\dev\olhos\nest-test\node_modules\file-type\package.json
    at exportsNotFound (node:internal/modules/esm/resolve:304:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:594:13)
    at resolveExports (node:internal/modules/cjs/loader:590:36)
    at Module._findPath (node:internal/modules/cjs/loader:667:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1129:27)
    at Module._load (node:internal/modules/cjs/loader:984:27)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (D:\dev\olhos\nest-test\src\app.controller.ts:3:1)
    at Module._compile (node:internal/modules/cjs/loader:1369:14) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
```

To simply reproduce the problem:

```
npm i -g @nestjs/cli
nest new nest-test
cd nest-test
npm i file-type
```

Edit the sample controller, scaffolded from nest.js:

```
import { Controller, Get } from '@nestjs/common';
import { AppService } from './app.service';
import { fileTypeFromFile } from 'file-type';

@Controller()
export class AppController {
  constructor(private readonly appService: AppService) {}

  @Get()
  async getHello(): Promise<string> {
    return (await fileTypeFromFile('Unicorn.png')).mime;
  }
}
```

`nest start`

Any idea?
Thanks

sindresorhus:
You need to open an issue on Nest.js instead. This package is configured correctly. They probably don't properly support the `exports` field in package.json.

RobertBeilich:
Coincidently had the same problem the same day, just a couple hours earlier :D Following up here to share what I've found:

The problem lies on neither side. It's Typescript, which compiles the dynamic import down to a `require`, which in turn obviously does not work.
Here are some options to circumvent it: https://github.com/TypeStrong/ts-node/discussions/1290

[...]

PR #665, opened 2024-08-28
Add CommonJS entry points

Borewit:
Just out of curiosity, as there are so many issues with legacy CommonJS projects, trying to import this pure ESM module.
Would this solution help a bit?

I added 2 CommonJS `require` entry points to the export:
1.  `core.cjs`,  CommonJS default entry-point
1. `node.cjs`, Node.js entry-points. 

Each entry point provides a function to load (using `loadFileType`) the ESM module asynchronous for CommonJS use.

```js
const { loadFileType } = require('file-type');

(async () => {
    const { fileTypeFromFile } = await loadFileType(); // Load the ESM file-type module async
    const type = await fileTypeFromFile('fixture.gif');
    console.log(type);
})();
```

Would this help?

Related to #661 and many more issues

### DIFF
--- a/core.cjs
+++ b/core.cjs
@@ -1,0 +1,7 @@
+/**
+ * CommonJS core (default) entry point
+ */
+
+module.exports = {
+	loadFileType: () => import('./core.js'),
+};
--- a/node.cjs
+++ b/node.cjs
@@ -1,0 +1,7 @@
+/**
+ * CommonJS Node entry point
+ */
+
+module.exports = {
+	loadFileType: () => import('./index.js'),
+};
--- a/package.json
+++ b/package.json
@@ -15,16 +15,19 @@
 		".": {
 			"node": {
 				"types": "./index.d.ts",
-				"import": "./index.js"
+				"import": "./index.js",
+				"require": "./node.cjs"
 			},
 			"default": {
 				"types": "./core.d.ts",
-				"import": "./core.js"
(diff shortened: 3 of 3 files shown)


sindresorhus:
I'm not interested in adding this.

Azbesciak:
@sindresorhus Sorry but I really do not get you.
The change you introduced 2 years ago thinking it is already common, is still no longer common 2 years later.
And IMHO it will not for the next 5 years, if not longer looking at the adoptation and all what is required to have it.
Simple thing - why should I add `.js` extension in every import, not to mention my files are `.ts`?
As long as it is not transparent from bundlers perspective, it will not happen.


If I need to decide - if I want some small library for detecting file type or to have MSSQL driver support, what I chose?
Will I migrate the whole big commercial app just to have one small library?
What do you think?
BTW `await import` does not work for me, in nestjs + typescript at least.

Borewit:
Workaround for loading **file-type** pure ESM module in TypeScript CommonJS project:

```ts
import * as path from 'path';
import {loadEsm} from 'load-esm';

/**
 * Import 'file-type' ES-Module in CommonJS Node.js module
 */
(async () => {
  const { fileTypeFromFile } = await loadEsm('file-type');

  const type = await fileTypeFromFile(path.join('..', 'fixture', 'fixture.gif'));
  console.log(type);
})();
```

PR #670, opened 2024-09-08
Improve Nikon RAW NEF (Tiff) format detection

### LINKED ISSUE #624: Detect NEF format, the result is image/tiff
NIKON D610, AF-S Nikkor 24-120mm f/4G ED VR

arraybuffer:

0x4D4D002A 00000008 001C00FE 00040000

Borewit:
Presumably resolves #624 

Some of the difference are caused by bytes being swapped, which looks like differences in [endianness](https://en.wikipedia.org/wiki/Endianness).

### DIFF
--- a/core.js
+++ b/core.js
@@ -1706,11 +1706,18 @@ export class FileTypeParser {
 					};
 				}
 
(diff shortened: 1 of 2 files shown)


Borewit:
Let's merge this one, probably able to detect for Nikon RAW NEF formats

PR #679, opened 2024-10-13
Add support for APK

epicmet:
Docs: https://developer.android.com/build/building-cmdline#package_pre-compiled_code_and_resources

As far as I understood, the `classes.dex` file is only specific for `apk` files and not other ext names (like .apks, .aab, .xapk, .apkm, .akp)

I don't know a simple "if statement" is good enough if you plan to support other ext names I mentioned (or even `.jar` or other zip file formats) But this PR should be sufficient enough for detecting only `.apk` files correctly.


Borewit (approved):
LGTM

PR #687, opened 2024-11-14
Add support for TTC

SethWen:
[See](https://learn.microsoft.com/en-us/typography/opentype/spec/otff#collections)


sindresorhus:
Wait, I just noticed that you submitted this against the wrong branch. You need to submit it against the `main` branch.

SethWen:
> Wait, I just noticed that you submitted this against the wrong branch. You need to submit it against the `main` branch.

Yes, I am currently using the commonjs(`16 branch`) version. Due to compatibility issues, my project cannot use the module(`main branch`) version.
I don't know if the `main branch` and the `16 branch` are functionally consistent? Is there a synchronization mechanism for these two versions, or will the 16 branch be deprecated?

PR #689, opened 2024-11-18
Return `pptx` type based on `/ppt*` path prefix occurrence in zipped files

### LINKED ISSUE #688: Detecting a pptx taking a very long time
### Description

Hello,

I have this library running in AWS via S3 in order to detect uploaded file types. I use fileTypeFromTokenizer.

When it came to a specific pptx file, it seemed to hang and take a very long time. Several minutes were not enough. 

I eventually got it to correctly detect the type after 7 minutes.

Looking at the code I could see it was going through and getting zip headers. This was a relatively large PPT (17MB) with more than 20 photos bundled into it. Big, but not abnormally big. 

Each one was downloaded like this:

```
{
  compressedSize: 631,
  uncompressedSize: 631,
  filenameLength: 16,
  extraFieldLength: 0,
  filename: '[trash]/0000.dat'
}
{
  compressedSize: 0,
  uncompressedSize: 0,
  filenameLength: 16,
  extraFieldLength: 0,
  filename: 'docProps/app.xml'
}
{
  compressedSize: 0,
  uncompressedSize: 0,
  filenameLength: 24,
  extraFieldLength: 0,
  filename: 'ppt/fonts/font23.fntdata'
}
... more fonts 
{
  compressedSize: 0,
  uncompressedSize: 0,
  filenameLength: 20,
  extraFieldLength: 0,
  filename: 'ppt/media/image1.png'
}
... more images
```

It finally got to 
```
{
  compressedSize: 0,
  uncompressedSize: 0,
  filenameLength: 33,
  extraFieldLength: 0,
  filename: 'ppt/notesMasters/notesMaster1.xml'
}
```

And then was able to see it ended in ".xml" and reported back the content type.

I really know very little about detecting Mime types, so forgive me if this is completely wrong, but I was wondering if the filename "ppt/*" was enough to declare the file type and shortcut downloading everything? Similar to a shortcut provided for xlsx? 

I will put up a PR if I can.

Thanks.

### Existing Issue Check

- [X] I have searched the existing issues and could not find any related to my problem.

- [X] My project is an [ESM project](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) and my `package.json` contains the following entry: `"type": "module"`.

- [X] I understand that file-type detects binary file types and not text or other formats.

Resolves #688

scottdoc:
Similar to how xlsx is shortcutted based on the path filename of the zipped headers, I have done the same for pptx.

Please feel free to reject if this is totally wrong and not a safe or acceptable way to detect pptx files. In my testing this commit made detetcting this file go from 7 mintues to a few seconds.

Resolves #688

### DIFF
--- a/core.js
+++ b/core.js
@@ -456,6 +456,13 @@ export class FileTypeParser {
 						};
 					}
 
+					if (zipHeader.filename.startsWith('ppt/')) {
+						return {
+							ext: 'pptx',
+							mime: 'application/vnd.openxmlformats-officedocument.presentationml.presentation',
+						};
+					}
+
 					if (zipHeader.filename.startsWith('3D/') && zipHeader.filename.endsWith('.model')) {
 						return {
 							ext: '3mf',


Borewit:
In many cases, the [tokenizer-s3](https://github.com/Borewit/tokenizer-s3) library can significantly enhance the speed of S3 file detection.

I am not in favor of relying solely on matching the `ppt/` path, as this approach feels more like an educated guess rather than a reliable method for identifying a `pptx` file.

A more robust solution might involve extracting and analyzing the `docProps/app.xml` file, which offers a much more specialized detection mechanism. However, this approach requires decompression and XML parsing, which adds complexity. Consequently, it might be better to delegate this functionality to an add-on for Office or zipped formats.

scottdoc:
Yes, I do use the tokenizer/s3 library.

My code is more or less:

```
import { fileTypeFromTokenizer } from "file-type";
import { makeTokenizer } from "@tokenizer/s3";

const tokenizer = await makeTokenizer(s3Client, {
			Bucket: location.bucket,
			Key: location.key
		});

const type = await fileTypeFromTokenizer(tokenizer);
await tokenizer.close();
```

Issue #692, opened 2024-11-22
Allowed return type for custom detectors are to narrow


joeried:
### Description

I have implemented a custom detector that also works without any problems.

However, the typescript types seem to be a bit too restrictive for the return value. Custom detectors must return a `FileTypeResult` (which limits them to already existing extensions/mine-types).

Am I mistaken, or shouldn't the `FileTypeResult` type actually be extendable somehow?

### Existing Issue Check

- [X] I have searched the existing issues and could not find any related to my problem.

### File-Type Scope Acknowledgment

- [X] I understand that file-type detects binary file types and not text or other formats.

Borewit:
I agree with your point. Furthermore, this issue extends beyond the types returned by custom detectors (as specified in the TypeScript typings). Since the core file-type library allows the addition of custom detectors, there’s no guarantee that the returned MIME types will align with the ones officially enumerated in the library’s typings. This introduces potential inconsistencies that developers need to handle when using custom detection logic.

joeried:
Please excuse the delay in replying, I have only just seen that there is a PR on this topic.

Shouldn't it theoretically be possible to infer the exact types via the return types of the `customDetectors` that are passed to the constructor? Am I missing something?

sindresorhus:
It may be possible, but the exact types are not that useful anyway, so no point in making it so complicated.

PR #693, opened 2024-11-23
Drop typed MIME-type and extension enumeration

Borewit:
This PR introduces a change where the MIME type (mime) and extension (ext) are no longer constrained by fixed enumerated TypeScript typings. Instead, these values are now represented as strings. This change has the following implications:

**Drops Fixed Enumerated Typings:**
Previously, the library’s TypeScript typings enforced that mime and ext could only be one of a predefined set of values. This change removes that strict enumeration, allowing greater flexibility, especially when custom detectors are involved.

**Supports Custom Detectors:**
The flexibility aligns with the implementation introduced in [#603](https://github.com/sindresorhus/file-type/pull/603), where custom detectors were added. Custom detectors can return MIME types and extensions beyond the predefined list, and enforcing enumerated typings would have conflicted with this capability.

**Resolves Issue #692:**
This change addresses the concerns raised in [#692](https://github.com/sindresorhus/file-type/issues/692), where the library’s core functionality and custom detector extensibility were at odds with the strict TypeScript typings. By treating mime and ext as strings, the library now accommodates both core-supported and custom-detected types without imposing unnecessary constraints.

**Benefits:**
Extensibility: Users can define custom detectors that return unique MIME types and extensions, even if these are not officially recognized by the core library.
Type Safety Tradeoff: While dropping fixed enumerated typings slightly reduces the strictness of TypeScript’s type safety, it ensures the library remains flexible and compatible with its extensibility features.
This approach maintains a balance between usability and flexibility, ensuring file-type can cater to diverse use cases without breaking custom implementations.


sindresorhus:
I agree

sindresorhus:
You need to update https://github.com/sindresorhus/file-type/blob/main/.github/pull_request_template.md too

PR #700, opened 2024-12-09
Add instructions on how to load file-type in a CommonJS + TypeScript

Borewit:
Related to:
-  #665

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -15,42 +15,14 @@ We accept contributions for commonly used modern file formats, not historical or
 ```sh
 npm install file-type
 ```
-
-> [!IMPORTANT]
-> This package is a **pure** ESM package, which means it can only be imported as an ESM module.
-> [Learn more about ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).
+**This package is an ESM package. Your project needs to be ESM too. [Read more](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).
+For TypeScript + CommonJS, see [`load-esm`](https://github.com/Borewit/load-esm)**
 
 If you use it with Webpack, you need the latest Webpack version and ensure you configure it correctly for ESM.
 
-### Importing in a CommonJS project
-
-We recommend updating your project to ESM for better compatibility and future-proofing.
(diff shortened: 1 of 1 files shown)


sindresorhus:
It's not clear from the readme how `load-esm` is any different from just using `await import()`?

Borewit:
Using `await import` in a CommonJS TypeScript project poses challenges because the TypeScript compiler transpiles `import()` statements to `require()` calls when module is set to CommonJS in tsconfig.json. This behavior conflicts with the dynamic nature of `import()` used for ESM.

Workarounds, such as wrapping the `import()` statement within `eval()` or similar constructs, can prevent TypeScript from transpiling it, but these approaches are clunky and error-prone.

The utility of [load-esm](https://github.com/Borewit/load-esm) lies in its simplicity: it bypasses the TypeScript compiler by executing the `import()` outside the compilation scope. By doing so, it maintains the intended behavior of `import()` for loading ESM modules dynamically, providing a clean and effective solution for such scenarios.

Borewit:
Should I better explain that in the README?

Borewit:
Sorry, I misread your question initially, improved the explanation the README @sindresorhus

sindresorhus:
I'm not interested in documenting how to use it with CommonJS output for TypeScript. That is very specific and every single package should not need to document this. I will accept a link to `load-esm` only because you are a maintainer here, even though I strongly disagree with the usage of this. The correct solution is to just output ESM.

readme.md
```diff
@@ -16,13 +16,18 @@ We accept contributions for commonly used modern file formats, not historical or
 npm install file-type
 ```
 
-**This package is an ESM package. Your project needs to be ESM too. [Read more](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).**
```
sindresorhus, line 19: This would be enough:

```
**This package is an ESM package. Your project needs to be ESM too. [Read more](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). And for TypeScript + CommonJS, see [`load-esm`](https://github.com/Borewit/load-esm)**
```

the rest can be documented in your load-esm readme.

PR #704, opened 2024-12-15
Give API access to  `FileTypeParser.detectors`

Borewit:
This gives the user more control to determine the sequence of detectors.

Does not break existing API.

Resolves: sindresorhus/file-type#628


core.d.ts
```diff
@@ -180,7 +180,10 @@ This method can be handy to put in a stream pipeline, but it comes with a price.
 export function fileTypeStream(webStream: AnyWebReadableStream<Uint8Array>, options?: StreamOptions): Promise<AnyWebReadableByteStreamWithFileType>;
 
 export declare class FileTypeParser {
-	detectors: Iterable<Detector>;
+	/**
+	File-type detectors
```
sindresorhus, line 184: Needs to be a bit more comprehensive. For example, that it contains the core detectors by default.

core.d.ts
```diff
@@ -180,7 +180,10 @@ This method can be handy to put in a stream pipeline, but it comes with a price.
...
+	 */
```
sindresorhus, line 185:
```suggestion
	*/
```

core.d.ts
```diff
@@ -180,7 +180,10 @@ This method can be handy to put in a stream pipeline, but it comes with a price.
...
+	readonly detectors: Detector[];
```
sindresorhus, line 186: `readonly` will prevent editing it, which is the point of this pull request.

sindresorhus (changes requested):

core.d.ts
sindresorhus: And changing from iterable to array *is* a breaking change. I'm fine with it though, but just pointing it out:

> Does not break existing API.

core.js
```diff
@@ -57,19 +57,18 @@ export async function fileTypeStream(webStream, options) {
 
 export class FileTypeParser {
 	constructor(options) {
-		this.detectors = options?.customDetectors;
+		this.detectors = options?.customDetectors ?? [];
```
sindresorhus, line 60: `customDetectors` could be an iterable, so it's not correctly handled here.

sindresorhus (changes requested):

core.d.ts
Borewit: I believe the `readonly` makes the property `detectors` immutable, but the content of the array can be changed. So new detectors can be added, or removed.

Borewit: The reason I put the `readonly` there is to making it a bit more difficult for the user to remove the default detectors. But it also limits the user in creating a new array of detectors. For maximum flexibility, I removed the `readonly`.

core.d.ts
```diff
@@ -114,11 +114,14 @@ console.log(await fileTypeFromBlob(blob));
 export declare function fileTypeFromBlob(blob: Blob): Promise<FileTypeResult | undefined>;
 
 /**
-Function that allows specifying custom detection mechanisms.
+A custom detector is a function that allows specifying custom detection mechanisms.
```
sindresorhus, line 117: This sentence does not actually say much.

```suggestion
A custom file type detector.
```

core.d.ts
```diff
@@ -180,7 +183,11 @@ This method can be handy to put in a stream pipeline, but it comes with a price.
 export function fileTypeStream(webStream: AnyWebReadableStream<Uint8Array>, options?: StreamOptions): Promise<AnyWebReadableByteStreamWithFileType>;
 
 export declare class FileTypeParser {
-	detectors: Iterable<Detector>;
+	/**
+	File-type detectors, initialized with a single entry holding the build in detector function.
+	The array is used to iterate over all detectors, as long as the function returns `undefined`.
+	*/
```
sindresorhus, line 189:
```suggestion
	/**
	File type detectors.
	
	Initialized with a single entry holding the built-in detector function.
	*/
```

sindresorhus: > The array is used to iterate over all detectors, as long as the function returns `undefined`.

This is not really relevant here. It can be explained if needed in the detector docs.

core.d.ts
Borewit: Simplify file type detector description

readme.md
```diff
@@ -344,7 +344,9 @@ A custom detector is a function that allows specifying custom detection mechanis
```
sindresorhus, line 344: This is out of sync now.

Borewit: It is out of sync, which I did not do in purpose, but syncing it with the `core.d.ts` does not make more readable.
They are not describing the exact same thing. `core.d.ts` is describing the custom detector type / interface, while the README is describing it having the header above as the only reference.

PR #707, opened 2024-12-18
Deprecate `NodeFileTypeParser`

Borewit:
In Node.js class `FileTypeParser` shall contain the Node.js specific functions (`fromFile()`) as well.

Why: This simplify the usage and documentation of the `FileTypeParser` class.

`NodeFileTypeParser` still works, as this is defined as alias for `FileTypeParser`.


Borewit:
I have no problem removing `NodeFileTypeParser` entirely, make it obsolete, elevating this change to a breaking API change.
I don't think the class `NodeFileTypeParser` is very widely used.

sindresorhus:
Just remove `NodeFileTypeParser`.

Issue #483, opened 2021-08-16
Allow disabling unsafe types?


rexxars:
There's a few file types that are marked as ["unsafe"](https://github.com/sindresorhus/file-type/blob/c037ba7ed901bd5efe57bbba707b607035265eaf/core.js#L1179) - and looking at the detection signatures I assume it's because the check is somewhat simple/naive. I have a gLTF file that is being detected as an `ico`, for instance.

Would it be possible to either:

a) Be able to disable the unsafe types (`fromBuffer(buf, {allowUnsafe: false})` or similar)
b) Return an `unsafe: true` in the return value for the unsafe types
c) Expose an array or similar that declares the unsafe types so we can exclude them on the consumer side

The issue with c) is that some of the unsafe file types does have more "safe" routes (like mpg).

Borewit:
I did not immediately understood what you meant with unsafe signatures, but I get it now, it are signatures beyond this point:

https://github.com/sindresorhus/file-type/blob/c037ba7ed901bd5efe57bbba707b607035265eaf/core.js#L1179

these are much more likely to be false positives.

rexxars:
> I did not immediately understood what you meant with unsafe signatures, but I get it now, it are signatures beyond this point:
> 
> https://github.com/sindresorhus/file-type/blob/c037ba7ed901bd5efe57bbba707b607035265eaf/core.js#L1179
> 
> these are much more likely to be false positives.

Yep, apologies for the vagueness. I've updated the original issue with a link.

Would you be willing to accept a PR? Any preferred approach?

midmarch:
I had a look to the MPEG-1 detection based on your feedback.

https://github.com/sindresorhus/file-type/blob/c037ba7ed901bd5efe57bbba707b607035265eaf/core.js#L1181-L1189

> some of the unsafe file types does have more "safe" routes (like mpg).

I don't think there is an alternative detection for that present in the code.

I actually tlooked into making it more safe. From the [non official specs](http://dvdnav.mplayerhq.hu/dvdinfo/mpeghdrs.html) it looks the current filter is actually already to narrow (at least on the bytes it is testing on). It's currently to specific on the stream-id. 

> a) Be able to disable the unsafe types (fromBuffer(buf, {allowUnsafe: false}) or similar)

I think it makes sense to optionaly exclude likely false positives (the so called _unsafe signatures_). I would prefere `includeUnsafe`.

> b) Return an unsafe: true in the return value for the unsafe types

[...]

PR #717, opened 2025-01-02
Separate confident and imprecise detectors, introduce detector identification

Borewit:
Extend the `Detector` type with an `id` property, which allows users to understand which detector is doing what. The core detectors are seperated in:
- `default.safe`: Safe core detectors
- `default.unsafe`: Collection of detectors with higher probability of false positive

Resolves: #483

### DIFF
--- a/core.js
+++ b/core.js
@@ -129,7 +129,9 @@ export async function fileTypeStream(webStream, options) {
 
 export class FileTypeParser {
 	constructor(options) {
-		this.detectors = [...(options?.customDetectors ?? []), this.parse];
+		this.detectors = [...(options?.customDetectors ?? []),
+			{id: 'core.safe', detect: this.detectCore},
+			{id: 'core.unsafe', detect: this.detectUnsafe}];
 		this.tokenizerOptions = {
 			abortSignal: options?.signal,
 		};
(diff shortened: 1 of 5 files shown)


sindresorhus:
Is "unsafe" the right word? It makes it sound like they have vulnerabilities, which is not correct.

Maybe: unreliable, unstable, imprecise, fragile, unstable

Borewit:
Loose, lax, speculative, unreliable?

sindresorhus:
Lets go with `fragile`.

core.js
```diff
@@ -129,7 +129,9 @@ export async function fileTypeStream(webStream, options) {
 
 export class FileTypeParser {
 	constructor(options) {
-		this.detectors = [...(options?.customDetectors ?? []), this.parse];
+		this.detectors = [...(options?.customDetectors ?? []),
+			{id: 'core.confident', detect: this.detectConfident},
+			{id: 'core.speculative', detect: this.detectSpeculative}];
```
sindresorhus, line 134:
```suggestion
			{id: 'core', detect: this.detectConfident},
			{id: 'core.fragile', detect: this.detectSpeculative}];
```

?

sly7-7: `uncertain`?

Borewit: Done

Issue #696, opened 2024-11-27
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in \node_modules\file-type\package.json


lalithasj:
### Description

Hi,

I got the following error without "types": "module":

node:internal/modules/cjs/loader:641
      throw e;
      ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in \node_modules\file-type\package.json     
    at exportsNotFound (node:internal/modules/esm/resolve:304:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:594:13)
    at resolveExports (node:internal/modules/cjs/loader:634:36)
    at Module._findPath (node:internal/modules/cjs/loader:724:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1211:27)
    at Module._load (node:internal/modules/cjs/loader:1051:27)
    at Module.require (node:internal/modules/cjs/loader:1311:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (E:\WebDevelopment\AngularWithNestJS\linkedin-clone\api\src\auth\helpers\image-storage.ts:5:1)
    at Module._compile (node:internal/modules/cjs/loader:1469:14) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v20.17.0

If I added "types": "module" error follows: 

/dist/main.js:2
Object.defineProperty(exports, "__esModule", { value: true });
                      ^

ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '\package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at file: /dist/main.js:2:23
    at ModuleJob.run (node:internal/modules/esm/module_job:234:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:473:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:123:5)

Node.js v20.17.0

Appreciate any suggestions.

### Existing Issue Check

- [X] I have searched the existing issues and could not find any related to my problem.

### ESM (ECMAScript Module) Requirement Acknowledgment

- [X] My project is an [ESM project](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) and my `package.json` contains the following entry: `"type": "module"`.

[...]

Issue #671, opened 2024-09-10
file-type ERR_PACKAGE_PATH_NOT_EXPORTED typescript


Shazam72:
Seems like file-type is not working well or to be exact can't be found in a Typescript environment. My config is bad is working pretty well with the other libraries in my initial project so i guess file-type is the problem here.
I isolated it in a separate project to see if the same problem occur again and it does.
This is the error I get:

```
[Typescript]
[Node] Error: No "exports" main defined in node_modules/file-type/package.json
[Node]     at exportsNotFound (node:internal/modules/esm/resolve:302:10)
[Node]     at packageExportsResolve (node:internal/modules/esm/resolve:592:13)
[Node]     at resolveExports (node:internal/modules/cjs/loader:596:36)
[Node]     at Function.Module._findPath (node:internal/modules/cjs/loader:673:31)
[Node]     at Function.Module._resolveFilename (node:internal/modules/cjs/loader:1135:27)
[Node]     at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename] (/test/node_modules/@cspotcode/source-map-support/source-map-support.js:811:30)
[Node]     at Function.Module._load (node:internal/modules/cjs/loader:990:27)
[Node]     at Module.require (node:internal/modules/cjs/loader:1237:19)
[Node]     at require (node:internal/modules/helpers:176:18)
[Node]     at/test/src/server.ts:2:12 {
[Node]   code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
[Node] }
[Node] ts-node src/server.ts exited with code 1
```
The steps to reproduce are simple. Here is my tree:
```
├── dist
│   └── server.js
├── package.json
├── package-lock.json
├── src
│   └── server.ts
├── test.jpg
└── tsconfig.json
```

in `src/server.ts`: 
```
(async ()=>{
   const fileType = await import("file-type")
   const type = await fileType.fileTypeFromFile("test.jpg")
   console.log(type)
})()
```

`package.json`:
```
{
    "name": "test",
    "version": "1.0.0",
    "main": "server.ts",
    "scripts": {
        "test": "echo \"Error: no test specified\" && exit 1",
        "dev": "concurrently -k -n \"Typescript,Node\" -c \"blue,green\" \"tsc --watch\" \"ts-node src/server.ts\""
    },
    "dependencies": {
        "@types/file-type": "^10.9.1",
        "file-type": "^19.5.0"
    },
[...]

Issue #668, opened 2024-09-05
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined


bokjo:
I'm doing an upgrade from ^16.5.3 to ^19.4.1  
(because ^16.5.3 stopped working on Node v22 for some reason, `fromBuffer` always gives undefined for file type...)

the current setup is:
Node: v22.6.0 (same outcome on v20.17.0 as well)
NextJS
module resolution CommonJS
tsconfig: `"module": "CommonJS"` and `"target": "ES2021"`

Im getting the following error: 

```sh
node:internal/modules/cjs/loader:646
      throw e;
      ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /Users/bojanche/.../node_modules/file-type/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:296:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:586:13)
    at resolveExports (node:internal/modules/cjs/loader:639:36)
    at Module._findPath (node:internal/modules/cjs/loader:747:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1234:27)
    at Module._load (node:internal/modules/cjs/loader:1074:27)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:217:24)
    at Module.require (node:internal/modules/cjs/loader:1339:12)
    at Hook._require.Module.require (/Users/bojanche/.../node_modules/require-in-the-middle/index.js:167:34) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v22.6.0
```

any idea what is causing this?

sindresorhus:
From the readme:

> **This package is an ESM package. Your project needs to be ESM too. [Read more](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).**

rosko:
That's what I ended up with:

```js
export const FileType = {} as typeof import('file-type');

(async () => {
    // biome-ignore lint/security/noGlobalEval: 🤷‍♂️
    Object.assign(FileType, await eval('import("file-type")'));
})();
```

Just don't use it immediately after importing `FileType`. However it's safe to use in some async code (e.g., request or event handler).

Borewit:
[...]

Issue #600, opened 2023-05-27
Unable to use plugin: Error [ERR_REQUIRE_ESM]: require() of ES Module


jstdk:
I am unable to use this plugin in my Typescript bot as it throws the following error on import:

`import { fileTypeFromBuffer } from 'file-type';
`
Then

`console.log(await fileTypeFromBuffer(bytes));
`

Leads to 

```
/home/jost/.nvm/versions/node/v18.16.0/lib/node_modules/ts-node/dist/index.js:851
            return old(m, filename);
                   ^
Error [ERR_REQUIRE_ESM]: require() of ES Module /home/jost/Code/simp/node_modules/file-type/index.js from /home/joost/Code/simp/models/messages_media.ts not supported.
Instead change the require of index.js in /home/jost/Code/simp/models/messages_media.ts to a dynamic import() which is available in all CommonJS modules.
  
```

On SO they advice to add the following to package.json, but that breaks my project altogether

`"type": "module"`

Thanks for any suggestions

sindresorhus:
https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

sindresorhus:
> "type": "module"

This is required to use this package.

jstdk:
@sindresorhus, I was a deep rabit hole, but it works now with "type": "module"

nabilfreeman:
Got here and found this issue because I just had to downgrade file-type 2 versions.

I think the usage stats on NPM tell you everything you need to know about how the authoritarian ESM migration is going.

`16.5.4` is the last CommonJS version of `file-type` 🤷‍♂️

| Version | Downloads (Last 7 Days) | Tag          |
|---------|-------------------------|--------------|
| 16.5.4  | 1,950,260               | version-16  |
| 18.5.0  | 167,986                 | latest       |

wyardley:
I'm seeing an issue where it will work with Typescript, but not a jest test -- I don't think the right fix there is to set module to true.

DanielWTE:
npm install file-type@16.5.4

worked for me

PR #722, opened 2025-01-16
fix: Fix require(ESM) for Node v22

mdorda:
Fixes for Node 22:
- https://github.com/sindresorhus/file-type/issues/696
- https://github.com/sindresorhus/file-type/issues/671
- https://github.com/sindresorhus/file-type/issues/668
- https://github.com/sindresorhus/file-type/issues/661
- https://github.com/sindresorhus/file-type/issues/600
- ... and similar

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,16 +15,19 @@
 		".": {
 			"node": {
 				"types": "./index.d.ts",
-				"import": "./index.js"
+				"import": "./index.js",
+				"default": "./index.js"
 			},
 			"default": {
 				"types": "./core.d.ts",
-				"import": "./core.js"
+				"import": "./core.js",
+				"default": "./core.js"
 			}
 		},
 		"./core": {
 			"types": "./core.d.ts",
-			"import": "./core.js"
+			"import": "./core.js",
+			"default": "./core.js"
 		}
 	},
 	"sideEffects": false,


sindresorhus:
This is not correct.

mdorda:
So what is the correct solution to make it work in Node 22? I tried to explain my reasons here: https://github.com/sindresorhus/file-type/issues/696#issuecomment-2595740520

I will be happy to fix it, but I do not know what you consider to be correct.

caldrjir:
@sindresorhus May I ask what the correct solution is?
I am facing the same problem and this seems to be ok for me

mdorda:
Maybe to use `require`? It works in Node 22 as well: https://github.com/sindresorhus/file-type/pull/723

Borewit:
> @sindresorhus May I ask what the correct solution is? I am facing the same problem and this seems to be okay for me

I'll try to answer your question. **file-type** is a pure ECMAScript Module (ESM). There is nothing wrong with the exports configuration that defines the ESM entry points. In fact, the exports might be slightly more complex than that of other ESM modules because it defines two entry points:
1. A default entry point
2. A Node.js entry point, which adds a few Node.js–specific functions.
Each entry point has its own typings corresponding to that entry point.

#### loading ESM module with Node `require`
If `require` in Node 22 claims to be able to load ESM modules, it should be able to load file-type. There is nothing missing in the `exports` for that use case.

#### loading ESM in CommonJS (CJS)
We recommend migrating your project to ESM. However, if for any reason you decide to stick with CommonJS, you can load ESM modules using a dynamic import. You might find that inconvenient, but as module maintainers we cannot build a module that is fully CJS-compliant without introducing other issues.

```js
async fucntion run() {
  // Load file-type module in CommonJS (CJS) module
  const {fileTypeFromStream} = await import('file-type');
}
````

#### loading ESM module in TypeScript / CommonJS (CJS)
There’s an extra complication here: the TypeScript compiler tends to transform dynamic imports into require calls, and since `require` doesn’t always load ESM modules correctly, this remains an issue.

Correct solutions would be:
1. Node’s `require` implementation fixes its ESM support.
2. The TypeScript compiler respects dynamic imports instead of converting them to require calls.
 
A workaround for these problems is to use the [load-esm](https://www.npmjs.com/package/load-esm) package:
```ts
import {loadEsm} from "load-esm";

type FileTypeModule =  typeof import('file-type');

async fucntion run() {
  // Load file-type module in TypeScript  / CommonJS (CJS) module
   const {fileTypeFromStream} = await loadEsm<FileTypeModule>('file-type');
}
````
This also works in Jest, I just added example for that here: https://github.com/sindresorhus/file-type/issues/661#issuecomment-2644981222

**Comment by mdorda** [comment]
We have already discussed it https://github.com/sindresorhus/file-type/pull/723
I do not think there is anything else to add. True is, that this package is the only one we cannot update. We use all other pure ESM packages without any problem. Just saying.

**Comment by Borewit** [comment]
I did some more investigation, and I think there is something we can do to make this work. Node.js introdecued new condition introduced [**"module-sync"**](https://nodejs.org/api/packages.html#conditional-exports), which seems tackle this issue.

```json
	"exports": {
		".": {
			"node": {
				"types": "./index.d.ts",
				"import": "./index.js",
				"module-sync": "./index.js"
			},
			"default": {
				"types": "./core.d.ts",
				"import": "./core.js",
				"module-sync": "./core.js"
			}
		},
		"./core": {
			"types": "./core.d.ts",
			"import": "./core.js",
			"module-sync": "./core.js"
		}
	}
```

@sindresorhus, any objection we reopen this PR, and use `module-sync` enabling Node.js to import file-type as an ESM module using require?

sindresorhus:
> "module-sync"
> The format is expected to be ES modules that does not contain top-level await in its module graph

I don't want to guarantee that.

Borewit:
> > "module-sync"
> > The format is expected to be ES modules that does not contain top-level await in its module graph
> 
> I don't want to guarantee that.

As far as I can see we do not have any top-level await in the module graph. Therefor I see no issue declaring that this is the case, and that file-type it totally fine to be loaded as an ES Module via require. I wish they made this obvious option possible years ago.

Note that most your ESM modules are exporting their entry points via `default`, where the assumption that these can be loaded via require/ESM, is already made by Node 22. 

3/4 file-type dependencies ([@tokenizer/inflate](https://github.com/Borewit/tokenizer-inflate), [strtok3](https://github.com/Borewit/strtok3), [token-type](https://github.com/Borewit/token-type)) I maintain, and these are not using top-level await. Number 4 is one you own, [uint8array-extras](https://github.com/sindresorhus/uint8array-extras), which neither has any top level await, and is already implicit `module-sync` compliant via `default`.

I like to have this one in very much, as I build my open source code components to be easy accessible and easy to use on as many platforms as possible. I really see no harm in this change, and if we experience issues, we always change course.

sindresorhus:
> As far as I can see we do not have any top-level await in the module graph. Therefor I see no issue declaring that this is the case, and that file-type it totally fine to be loaded as an ES Module via require. I wish they made this obvious option possible years ago.

None now, but could be in the future. I don't want to be held back by `require`. Declaring this also means we are on the hook to deal with it if any dependency at any level starts using top-level await (which does not generally require a major version, btw) at any point in the future.

> easy to use on as many platforms as possible

All modern platforms support ESM.

---

It seems like this is an issue with Nest.js and it's not something we should have to deal with: https://github.com/nestjs/nest/issues/13319#issuecomment-2141110188

Borewit:
There is always a risk that dependency breaks our code. And there are plenty of ways to deal with that. 
But you cannot build software today on something what is planned to to be build to tomorrow.

> THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND

It does harm not if we allow esm-require for the time being, while the world is still very slowly migrating to ESM. It it is totally up to use for how long and under which condition we support that. But for this moment I am committed to do so.

imaksp:
Hi all, I believe this is good change by Node, and it will help so many older projects who can't migrate to full ESM easily. so please consider adding its support (as it seems top level await is not being used). Thanks.

PR #708, opened 2024-12-19
Document third-party file-type detectors

Borewit:
Adds the following to README:
- Refer to `@file-type/xml` for XML based formats
- Use `@file-type/xml` to demonstrate adding a custom detector

PR #724, opened 2025-01-26
Add support for SVG

Shashee99:
This PR adds support for detecting the SVG file format. Below are the details:

- **File format**: SVG (Scalable Vector Graphics)
- **MIME type**: `image/svg+xml`
- **Detection confidence**: `detectConfident()` — SVG files are identified with high certainty based on their distinct structure and signature.

### Changes made:
1. **Fixture file**: Added `fixture.svg` to the `fixtures` directory.
2. **Supported file types**: 
   - Added `.svg` extension to the `extensions` array in `supported.js`.
   - Added MIME type `image/svg+xml` to the `types` array in `supported.js`.
3. **Detection logic**: Added the logic for detecting SVG files in the `core.js` file.
4. **Confidence category**: Since the structure of SVG files is unique and easy to detect, this detection is classified under `detectConfident()`.
5. **Documentation**:
   - Updated the "Supported file types" section of the README to include SVG in alphabetical order.
   - Added `.svg` extension to the `keywords` array in the `package.json`.

### References:
- **File format specification**: [SVG - W3C](https://www.w3.org/TR/SVG2/)


sindresorhus:
From the top of the readme:

> This package is for detecting binary-based file formats, not text-based formats like .txt, .csv, .svg, etc.

There are also many things wrong with this pull request, and I also suspect it's AI generated based on the pull request description.

Shashee99:
Thank you for your feedback! I understand now that this package is focused on detecting binary-based file formats. However, I added support for SVG because it is a common and well-defined format, and it includes a structured text format that is easy to detect programmatically (via its unique XML-based signature). While SVG is technically text-based, it’s used widely in many image processing workflows, which is why I thought it could be a useful addition.

Borewit:
I wrote a file-type plugin for XML file types, which does detect SVG as well: [@file-type/xml](https://github.com/Borewit/file-type-xml), which would be nice to mention somewhere in the README, as proposed in PR #708.

PR #730, opened 2025-02-01
Add integration tests with Bum (JavaScriptCore)

Borewit:
Add integration for [Bum](https://bun.sh/) 1.2. Bum is based on the [JavaScriptCore](https://github.com/apple-opensource/JavaScriptCore) (which is also used in the Apple Safari web browser).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,8 +3,8 @@ on:
   - push
   - pull_request
 jobs:
-  test:
-    name: Node.js ${{ matrix.node-version }}
+  test-nodejs:
+    name: Node.js (V8) ${{ matrix.node-version }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't really see the need for this. Bun promises Node.js compatibility. For cases where this is not working, it's up to Bun to fix it, not us. So I think this change will just slow us down.

Borewit:
My actual objective was actually not be Bum complaint, but to be compatible with modern commonly used web browsers, which in view are: 
1. Chromium (Chrome & Edge)
2. Firefox 
3. Apple Safari

The dependencies [strtok3](https://github.com/Borewit/strtok3) and [peek-readable](https://github.com/Borewit/peek-readable) used to require a [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) that supports a [ReadableStreamBYOBReader](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamBYOBReader). Same for the [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob) where the [stream()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream) should return such [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).

The issue is that Apple Safari does not support the BYOB flavor, so recently I started to implement fallbacks to [ReadableStreamDefaultReader](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader). But I do have a challenge testing this properly as do not have a way to run Safari. 

As believe this is the only issue remaining of being Node.js / browser cross compatible, I thought it would nice to address that one.

This PR this actually not meet the objective at all:
1. The tests are still run by Node.js, so the results have no value
2. Bun does support BYOB, so it not the right engine to mimic the Safari behavior

PR #736, opened 2025-02-09
Allow `require` to load file-type as ES Module

Borewit:
Add [`module-sync`](https://nodejs.org/api/packages.html#conditional-exports) in exports, allowing [require-esm loading](https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js) **file-type** with Node.js ≥ 22.
 
Discussion: https://github.com/sindresorhus/file-type/pull/722#issuecomment-2645592908

Superseding:
- #722
- #723. 

Related:
- #665
- #696
- #671
- #668
- #661
- #600

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,16 +15,19 @@
 		".": {
 			"node": {
 				"types": "./index.d.ts",
-				"import": "./index.js"
+				"import": "./index.js",
+				"module-sync": "./index.js"
 			},
 			"default": {
 				"types": "./core.d.ts",
-				"import": "./core.js"
+				"import": "./core.js",
+				"module-sync": "./core.js"
 			}
 		},
 		"./core": {
 			"types": "./core.d.ts",
-			"import": "./core.js"
+			"import": "./core.js",
+			"module-sync": "./core.js"
 		}
 	},
 	"sideEffects": false,


imaksp:
good, waiting for this, require esm support will also land in Node 20 LTS soon.

Borewit:
As pointed out here: https://github.com/sindresorhus/file-type/pull/737#issuecomment-2662501298

### 9 out of 10 of our users are still CommonJS users
Looking at the statistics of the last 7 days, **89.2%** of the downloads are still CommonJS and just **10.8%** are ESM versions. That tells me, despite ESM is around for many years, there is still a significant demand from CommonJS projects. 

With this PR we likely closing the gap for many, as there is clear bottleneck on the latest CommonJS version, which is still the most frequently downloaded version by far.

Borewit:
@sindresorhus, in reply to https://github.com/sindresorhus/file-type/pull/741#issuecomment-2674094436


> We can use await import inside the function.

This is a misunderstanding Sindre. It is certainly also possible to use [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) with ["module-sync"](https://nodejs.org/api/packages.html#conditional-exports). The restriction is:
> The format is expected to be ES modules that does not contain top-level await in its module graph

So this is not allowed:
```js
// ESM Entry point file

// Top level await
export quoteOfTheDay = await fetch('https://www.forbes.com/forbesapi/thought/uri.json?enrich=true&query=1&relatedlimit=1');
```

But this is totally fine:
```js
// ESM Entry point file

export async function detect(buffer {
  const {fileTypeFromBuffer} = await import('file-type'); // dynamic import
  return fileTypeFromBuffer(buffer );
}

sindresorhus:
> This is a misunderstanding Sindre.

That is completely unrelated to this issue. There I'm talking about using `await import` to be able to import `node:fs`.

Borewit:
Merged in #741

PR #739, opened 2025-02-20
Update readme.md

dioptre:
esm is unnecessary


sindresorhus:
It's better described in the "read more" link.

Borewit:
The current `exports` is blocking `file-type` to be loaded Node version 22, using `require`. PR #736 unblocks that.

PR #740, opened 2025-02-20
Add support for RealMedia

matt-oakes:
References:

- Wikipedia: https://en.wikipedia.org/wiki/RealMedia
- Header magic number: https://wiki.multimedia.cx/index.php?title=RealMedia#RealMedia_file_header_(.RMF_or_.RMP)

I have only added support for the RealMedia format and not the older RealAudio. This is because `ffmpeg` only supports this format and I don't have good fixtures for the older format.


Borewit (changes requested):
As per guidance:

> Existing signatures requiring same sample length (same signature group) will be tested prior to your new detections. Yours will be last. (rational: common formats first).

Please make this detection the last detection in group of 4 byte signatures.

matt-oakes:
@Borewit Thanks for the review. Sorry, I missed that part of the requirements.

I have fixed that now and put it at the bottom of the 4-byte section. I have also added the missing heading comment for the 5-byte `DRACO` check to make it clearer that isn't a 4-byte one. I can remove that from this PR if you prefer.

Borewit (approved):
Look good to me

matt-oakes:
Thank you for the review and merging this! Roughly how often do you produce a new release for this to be available publicly?

matt-oakes:
Thank you and thanks for your work on this library. It's very useful!

PR #741, opened 2025-02-21
Add './node' subpath export for node entry point

### LINKED ISSUE #652: Can't access node functions when tsconfig moduleResolution is set to "bundler"
I'm using tsup to bundle my node project into a single file, and have `moduleResolution: "bundler"` set in my tsconfig because a) it feels like the right choice as I'm using a bundler, and b) it allows me to ignore writing out file type extensions with esm. 

Unfortunately it looks like file-type only exposes node functions like `fileTypeFromFile` when moduleResolution is set to "node". I get that this is done to add extra safety and prevent consumers from importing functions that don't work in the environment that they are running, and 99% of projects with `moduleResolution: "bundler"` are likely headed for the browser, but considering that bundling is valid for node projects as well, I wonder whether the exports should be relaxed for this?

### LINKED ISSUE #738: fileTypeFromFile not found
### Description

```ts
import { fileTypeFromFile } from 'file-type'

'"file-type"' has no exported member named 'fileTypeFromFile'. Did you mean 'fileTypeFromBlob'?ts(2724)
```

### Existing Issue Check

- [x] I have searched the existing issues and could not find any related to my problem.

- [x] My project is an [ESM project](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) and my `package.json` contains the following entry: `"type": "module"`.

- [x] I understand that file-type detects binary file types and not text or other formats.

Borewit:
Add `./node` subpath export

`./core` has been inherited at the days this package was CJS, where the default exports was Node.js. The core allowed to import the functionality, which did not depend on the Node API. With ES Module implementation, it is the other way around. The non-Node is now the default. Therefor it make more sense expose the Node specific functionality, for those cases where conditional exports cannot be used. Such as the ``"moduleResolution": "bundler"` in TypeScript projects.

I chose to use the `default` export in this case  _as generic fallback that always matches_, as the subpath import is the user overruling the import in cases where the conditional imports have no, or not the desired effect.  

Fixes #652
Fixes #738

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,16 +15,19 @@
 		".": {
 			"node": {
 				"types": "./index.d.ts",
-				"import": "./index.js"
+				"import": "./index.js",
+				"module-sync": "./index.js"
 			},
 			"default": {
 				"types": "./core.d.ts",
-				"import": "./core.js"
+				"import": "./core.js",
+				"module-sync": "./core.js"
 			}
 		},
 		"./core": {
 			"types": "./core.d.ts",
-			"import": "./core.js"
+			"import": "./core.js",
+			"module-sync": "./core.js"
 		}
 	},
 	"sideEffects": false,
--- a/package.json
+++ b/package.json
@@ -26,8 +26,11 @@
 		},
 		"./core": {
 			"types": "./core.d.ts",
-			"import": "./core.js",
-			"module-sync": "./core.js"
+			"default": "./core.js"
+		},
+		"./node": {
+			"types": "./index.d.ts",
+			"default": "./index.js"
 		}
 	},
 	"sideEffects": false,


sindresorhus:
I think the separation of Node.js vs the rest has caused us way too many problems. How about we always export `fromFile`, but throw if it the `fs` API is not available. We can use `await import` inside the function.

That will probably fix all the other import issues too, since our package.json could go back to:

```
"exports": {
	"types": "./index.d.ts",
	"default": "./index.js"
}
```

Borewit:
https://github.com/sindresorhus/file-type/pull/736#issuecomment-2674284940

> I think the separation of Node.js vs the rest has caused us way too many problems. How about we always export fromFile, but throw inside the function if the fs API is not available. 

That will also mix the dependency tree, definitely a no go. I think in separation is  actually working great, and massive advantage of being the ECMAScript module.

Only in exceptional cases, this does not work, such as the ``"moduleResolution": "bundler"` in TypeScript case.

Yet there are workaround on the user side possible as well, such as raising the `node` condition manually:
```
{
  "compilerOptions": {
    "target": "es2022",
    "moduleResolution": "bundler",
    "customConditions": ["node"]
  }
}
```

> That will probably fix all the other import issues too, since our package.json could go back to:
> 
> "exports": {
> 	"types": "./index.d.ts",
> 	"default": "./index.js"
> }

Getting back to a more basic `exports`, I propose the following:
```json
	"exports": {
		"node": {
			"types": "./index.d.ts",
			"default": "./index.js"
		},
		"default": {
			"types": "./core.d.ts",
			"default": "./core.js"
		}
	},
```
- Drops the subpath exports.
- Uses `default` as fallback, shall work for require-esm
- Preserves the Node.js API with default, seperation

Yet I have nothing against the subpath exports, if that helps some users, it does not disturb anything else.
And that is basically what this PR does.

The biggest issue we have now are `import` conditions, without any fallback to default or module-sync.

sindresorhus:
> That will also mix the dependency tree

The same dependencies are installed for all export types, so not really sure what you mean.

Borewit:
> > That will also mix the dependency tree
> 
> The same dependencies are installed for all export types, so not really sure what you mean.

For example `node:fs` is only imported via the Node.js specific entry point. That Node specific API dependency would end-up in the (browser) default entry point, following your proposal.

sindresorhus:
The method would, yes, but we would simply make it throw when called and the `fs` api is not available. And as I already commented, we would use dynamic import inside the function, so it should have no effect on non-Node.js consumers, other than a few LOCs of unused code.

Borewit:
> The method would, yes, but we would simply make it throw when called and the `fs` api is not available. And as I already commented, we would use dynamic import inside the function, so it should have no effect on non-Node.js consumers, other than a few LOCs of unused code.

To me it feels going from one extreme to the other.

Again, I don't see why we should throw a good working mechanism overboard. Only in the TypeScript bundler the Node condition is not picked up, which probably relies in more in the nature of the bundler, or maybe even on the bundler project is setup. I see no reason to drastically change our architecture, but more then happy to add an extra sub-path export if that helps them to plugin the Node functionality. 

What is much more important, is to support require-esm, there is massive need for that solution.   

I propose to give this PR a try, and see how that works out. I have no problem with sustaining the `/core` subpath neither, to avoid any breaking, although I cannot imagine it still serves a useful purpose, but maybe I overlooked something.

sindresorhus:
> To me it feels going from one extreme to the other.

I don't really see what's extreme about it. There is really nothing inherently tied to Node.js for this package other than the file access, which can easily be isolated to a single method. Regardless of the sub-export problems, having the package as platform agnostic as possible is a good goal.

Borewit:
> Regardless of the sub-export problems, having the package as platform agnostic as possible is a good goal.

I totally agree with that! 

> I don't really see what's extreme about it. 

The contradiction I see is objecting to my proposal of extending the current import condition with `module-sync` while then proposing to use `default` instead—which effectively matches any condition. The current `import` condition limits us to **fewer** use cases (for instance, [require-esm loading](https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/) works but is now effectively prohibited), whereas `default` would open the door to **more** import scenarios, including options we do not support. Supporting _require-esm loading_ significantly improves accessibility for many users of this ESM module, so I believe we should take advantage of that opportunity. This approach addresses a major weak point in the CommonJS-to-ESM transition.

The second contradiction is moving away from a purist Node.js ESM specification. Your proposal replaces the clear-cut ESM conditions for accessing the Node API with a mechanism that relies on the success of lazy loading the Node.js API.

The third contradiction is the rejection of workarounds. Exporting a sub-path is a very minor change that doesn’t impact anything else, yet it’s dismissed solely because it’s seen as a workaround. In contrast, replacing the Node ESM condition with lazy loading is a drastic workaround with potential side effects.

sindresorhus:
> The second contradiction is moving away from a purist Node.js ESM specification.

Sub-exports is a Node.js invented thing and not part of ESM.

> Your proposal replaces the clear-cut ESM conditions for accessing the Node API with a mechanism that relies on the success of lazy loading the Node.js API.

No, both are completely valid methods to achieve the same goal.

> Exporting a sub-path is a very minor change that doesn’t impact anything else

I may seem minor, but it too can have side-effects. What if a package depends on it, and then a consumer of that package wants to use that package in the browser. Then that package has locked into using the Node.js API since they forced the `node` sub-export. Enabling users to shoot them self in the foot too easily is not good either.

> In contrast, replacing the Node ESM condition with lazy loading is a drastic workaround with potential side effects.

This change too has potential side-effects.

sindresorhus:
I don't really want to waste more time discussing this, so fine, we can do it your way.

package.json
```diff
@@ -15,16 +15,18 @@
...
+				"module-sync": "./core.js"
 			}
 		},
-		"./core": {
-			"types": "./core.d.ts",
-			"import": "./core.js"
+		"./node": {
+			"types": "./index.d.ts",
+			"default": "./index.js"
 		}
```
sindresorhus, line 30: Not worth doing a breaking change for this. So just add `node` and keep `core`, and we can remove `core` in the next major version.

Issue #582, opened 2023-02-15
.pdf files can be detected as .ai based on content


eric-yuan-vanta:
When pdf files have images created from photoshop or adobe ai in them, `file-type` detects them as .ai based on the byte checking heuristic we have in place. 

I'm proposing that even if the magic string is found, if the original file's extension is `.pdf`, `file-type` should consider it a pdf and not change it's type based on some content inside of it.

An even more strict approach that I would also support is only returning ai file type if the file extension is already `.ai`. It seems more natural/compatible to default to .pdf if .ai isn't explicitly specified, since the ai detection is just a loose heuristic anyway.

eric-yuan-vanta:
@sindresorhus Curious if you have thoughts on this. 

I plan to put up a fix for the second one, but I would like to get https://github.com/sindresorhus/file-type/pulls in first

Borewit:
None of the file implemented recognition is perfect (guaranteed to be correct). By writing 4 characters at the beginning of a text file you can probably mimic half of of the file recognition heuristics. And the reliability of heuristics vary.

If the recognition is likely to introduce false positives (for which there is no clear definition), it may indeed be better to, preferably improve the algorithm, or, like you suggest, fall back on it's parent file type.

Borewit:
We indeed can only determine the file-type based on the file content.
It helps if you make examples available, so we can analyze the difference. (Preferably small file sizes).

Borewit:
I fixed a bug (#728) close to this `.ai` (Adobe Illustrator) detection portion in the code. I don't trust this `.ai` detection algorithm at all. I am in favor of cutting it out. Either we do it properly, or we don't.

Although PDF is not a pure text based formats, I believe it requires text processing for better interpretation of the sub-format, which is arguable in scope of file-type, being focused on binary formats.

PR #743, opened 2025-02-25
Drop Adobe Illustrator (.ai) detection support

Borewit:
Drop Adobe Illustrator (`.ai`) file detection support as the detection mechanism is very poor quality (as raised in #582).

Skipping 1350 bytes, followed by a search in the next 10kB for a keywords is not a reliable mechanism.
There is no fixed offset at 1350 bytes, as well the 10kB is not based on any meaningful foundation.
And neither the keywords necessary from from a context which reliably indicates this is an Adobe Illustrator file.

Second objection I have, is that this is based on (poor) text file parsing, as PDF is a text based format.
Text based formats are not in the scope of file-type.

Reverses: #323
Resolves: #582

Related:
- https://github.com/sindresorhus/file-type/issues/582#issuecomment-2636569875


sindresorhus:
You could maybe consider AI detection for your XML detection library as I think I remember that you can detect Illustrator files by looking at its XMP metadata, which is XML.

Borewit:
> Agreed, but this is a breaking change, and we just did a major version, so I think this should wait a bit.

No problem, I am not in a rush.

> You could maybe consider AI detection for your XML detection library as I think I remember that you can detect Illustrator files by looking at its XMP metadata, which is XML.

Something like that. Starting with basic PDF decoding which I could maybe utilize [read-next-line](https://github.com/Borewit/read-next-line) to iterate over the lines. I am not very familiar with the PDF file format though.

Borewit:
Adobe Illustrator (`.ai`) detection has been implemented in [@file-type/pdf](https://github.com/Borewit/file-type-pdf)

Issue #746, opened 2025-04-24
Support for ppsx


PRR24:
### Description

To cover full set of application/vnd.openxmlformats-officedocument.*, please add a support for
.ppsx / application/vnd.openxmlformats-officedocument.presentationml.slideshow

Detection should be the exact same as for all other vnd.openxmlformats, only mapping seems to be missing.

### Existing Issue Check

- [x] I have searched the existing issues and could not find any related to my problem.

### File-Type Scope Acknowledgment

- [x] I understand that file-type detects binary file types and not text or other formats.

Borewit:
Are you able to provide a small sample file @PRR24 ?

PRR24:
Here you are. Attaching .ppsx and its corresponding macro-enabled counterpart .ppsm.

[pps.zip](https://github.com/user-attachments/files/19894849/pps.zip)

PR #747, opened 2025-04-25
Add support Office PowerPoint 2007 (macro-enabled) slide show

Borewit:
Normalized MIME-types to lowercase.
MIME-type are defined to be case insensitive.

Add support for:
- `.ppsx`, `application/vnd.openxmlformats-officedocument.presentationml.slideshow`
- `.ppsx`, `application/vnd.ms-powerpoint.slideshow.macroEnabled.12`

MIME type reference: [Register the 2007 Office system file format MIME types on servers](https://learn.microsoft.com/en-us/previous-versions/office/office-2007-resource-kit/ee309278%28v=office.12%29)

Resolves sindresorhus/file-type#746


PRR24:
Not related to the task, but reviewing the changes I noticed that all macroEnabled.12 mime-s are in camelCase except
```application/vnd.ms-excel.template.macroenabled.12```

package.json also got unrelated "packageManager" entry. Not sure, if it is intentional or a mistake.

Other than that, looks exactly as expected, nice job.

Borewit:
Thx! @PRR24  

- Removed `packageManager`, was not intentional .
- MIME are defined to be case insensitive. Adjusted other types be lower case, and the switch case insensitive.

PR #748, opened 2025-05-07
fix: official IANA MIME type for Apache Parquet files

JoeCap08055:
The MIME type 'application/x-parquet' is old/deprecated/unofficial. The official MIME type is: `application/vnd.apache.parquet`.
Also added a missing magic number for this type.

https://www.iana.org/assignments/media-types/application/vnd.apache.parquet

- **One PR per file type.**

### DIFF
--- a/core.js
+++ b/core.js
@@ -920,7 +920,7 @@ export class FileTypeParser {
 			};
 		}
 
-		if (this.checkString('PAR1')) {
+		if (this.checkString('PAR1') || this.checkString('PARE')) {
 			return {
 				ext: 'parquet',
 				mime: 'application/vnd.apache.parquet',
--- a/core.js
+++ b/core.js
@@ -923,7 +923,7 @@ export class FileTypeParser {
 		if (this.checkString('PAR1')) {
 			return {
 				ext: 'parquet',
-				mime: 'application/x-parquet',
+				mime: 'application/vnd.apache.parquet',
 			};
 		}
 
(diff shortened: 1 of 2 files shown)


JoeCap08055:
> Hi @JoeCap08055, thanks for correcting the MIME type.
> 
> I failed to find any trace of documentation mentioning `PARE` is used as a magic byte signature for Apache Parquet files. Can you clarify where this comes from and why we should include that signature? Do you have an **original** file using that signature you can share?

@Borewit I saw it in the official IANA descriptor [here](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet#:~:text=2%2E%20Magic%20number%28s%29%3A%20PAR1,%20PARE)

> Additional information:
>
> 1. Deprecated alias names for this type: x-parquet
> 2. **Magic number(s): PAR1, PARE**
> 3. File extension(s): .parquet
> 4. Macintosh file type code: not available
> 5. Object Identifiers: not available

Borewit:
Thanks for providing the background @JoeCap08055 , makes total sense to include that one indeed.

Borewit (approved):
LGTM

sindresorhus:
Looks good to me, but this is a breaking change and must wait for the next major version.

JoeCap08055:
> Looks good to me, but this is a breaking change and must wait for the next major version.

@sindresorhus that's unfortunate, but makes sense--is there a rough estimate of when that's expected?

Borewit:
@sindresorhus , we now have a number PR's with some minor API changes, I think this is a good moment to start merging those and start to work towards a major release.

Issue #750, opened 2025-05-12
False negative on some MP3s


myndzi:
### Description

Hi there. First off, kudos on great work here :) Really helpful library and does just what I want and no more.

I was using this library as part of an exploratory side project and noticed some MP3 files that didn't get identified.

I dug into it to see where the mismatch is, as well as comparing outputs of `file`, `ffprobe`, and `music-metadata`; all of these recognize it (but in different ways, by different mechanisms).

- ffprobe had a warning log that was a clue: `[mp3 @ 0x139705ac0] Skipping 10 bytes of junk at 62401.`
- `file` seems to be somewhat imprecise in its output (as noted in another issue) but can identify the content-type suitably
- `music-metadata` does return seemingly-correct data, which confused me at first because this library is a dependency of that one! But it turns out it just does a file-extension check before attempting to guess the format by content. Inspecting the returned object in a repl also shows about a thousand "empty tag" warnings suggesting the file in question is definitely a bit wack.
 
As for possible improvements, enabling the `debug` logs for `music-metadata` shows that when it actually parses the file it is doing it robustly, syncing to the frame sync markers*, and reading correct data after. So there's a possibilty that if you're just trying to _identify_ an mp3 (and not read the contents), that it's enough to attempt to sync the stream even if the first frame doesn't show up where it's expected after the ID3 tag (in this file at least, the ID3 tag is first)

\* It's been a while, I don't remember how MP3 sync works...

`libavformat` seems to use a heuristic, since it's not going in with the expectation that "this is an mp3 and I'm attempting to extract data from it" (which is what `music-metadata` seems to be doing from the extension check). `libavformat` scans over 64kb from its starting point and attempts to decode each position as the start of a frame; then, it extracts the frame length and checks if there is another valid frame at the expected position: https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/mp3dec.c#L419-L431

`file` is incomprehensible to me, and I can't even find where it defines the checks for actual MP3 files 🤷

I'm not sure if these approaches are within the scope of what you want for this library, since they are a bit "meaty" compared to simple magic checks. I can potentially submit a PR, but wanted to check the temperature before investing the time. What do you think?

### Existing Issue Check

- [x] I have searched the existing issues and could not find any related to my problem.

### File-Type Scope Acknowledgment

- [x] I understand that file-type detects binary file types and not text or other formats.

myndzi:
[...]

PR #751, opened 2025-05-12
Improve support for erroneous ID3v2 headers

myndzi:
After reading an ID3v2 header, if the file was unable to be identified, try again 10 bytes further in. See https://github.com/sindresorhus/file-type/issues/750


myndzi:
I kept digging and am happy that this solution is about as minimal as it can get. I totally understand if you don't want to merge it though; after all, it's a workaround for _incorrect_ data, and the proper solution is to fix the data, not the code.

There may however be value to adding it anyway - or even going a bit further with one of the "looser" options mentioned in the issue. Mp3 is so widespread, and has a long history of many different tools, encodings, and such -- it seems valuable both to improve detection (due to "market share") and because the landscape is likely littered with all sorts of broken data.

Borewit:
I think PR  #646 is a better solution to resolve out-of-sync MPEG data, as it handles multiple out-of-sync scenario's, not just the 10 bytes.

myndzi:
Yep, me too. The original issue was more of a query on whether you wanted to go that hard :) This is a very precise check for a specific error, so performance-wise it doesn't carry the same penalty as scanning an entire stream for sync bytes.

Close if you like, or if you want to accept it as a momentary improvement (since that other PR is 10 months old and still in draft), it's here 👍

Borewit:
Superseded by #646

PR #735, opened 2025-02-07
[WIP] make js / d.ts files consistent

sly7-7:
It's a WIP related #715

As per https://github.com/sindresorhus/file-type/issues/715#issuecomment-2596421250 , I've removed all the FileTypeOptions arguments passed in helper functions `fileTypeFromXXX`.
Some tests are now failing, it emphasizes the fact that some functions currently have the ability to take a FileTypeOptions argument but not others.

I've also reorganized a bit the file structure of core.js/core.d.ts and index.js/index.d.ts  to make them more or less aligned.

Please let me know if that makes sense, and what would you do about the "helpers" functions, make all of them or none of them able to take FileTypeOptions argument.

PR #752, opened 2025-05-15
Allow file-type options to be directly passed to exported functions

Borewit:
Changes:

- Allow file-type options to be directly passed to exported functions
- Improve coverage of type tests
- Centralized file-type options documentation (I want to introduce a new shared option in #646)

Related to #735, as this addresses some of the inconsistencies as well

### DIFF
--- a/core.d.ts
+++ b/core.d.ts
@@ -37,7 +37,7 @@ If file access is available, it is recommended to use `.fromFile()` instead.
 export function fileTypeFromBuffer(buffer: Uint8Array | ArrayBuffer, options?: FileTypeOptions): Promise<FileTypeResult | undefined>;
 
 /**
-Detect the file type of [web `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
+Detect the file type of a [web `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
 
 The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
 
--- a/core.d.ts
+++ b/core.d.ts
@@ -24,38 +24,38 @@ export type FileTypeResult = {
 };
 
 /**
-Detect the file type of a `Uint8Array`, or `ArrayBuffer`.
+Detect the file type of `Uint8Array`, or `ArrayBuffer`.
 
 The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
 
 If file access is available, it is recommended to use `.fromFile()` instead.
 
 @param buffer - An Uint8Array or ArrayBuffer representing file data. It works best if the buffer contains the entire file. It may work with a smaller portion as well.
-@param options - Options to override default behaviour.
+@param options - Options to override default behavior.
 @returns The detected file type, or `undefined` when there is no match.
 */
 export function fileTypeFromBuffer(buffer: Uint8Array | ArrayBuffer, options?: FileTypeOptions): Promise<FileTypeResult | undefined>;
 
 /**
-Detect the file type of a [web `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
+Detect the file type of [web `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
 
 The file type is detected by checking the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files) of the buffer.
 
 @param stream - A [web `ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) streaming a file to examine.
-@param options - Options to override default behaviour.
+@param options - Options to override default behavior.
 @returns A `Promise` for an object with the detected file type, or `undefined` when there is no match.
 */
 export function fileTypeFromStream(stream: AnyWebByteStream, options?: FileTypeOptions): Promise<FileTypeResult | undefined>;
 
 /**
-Detect the file type from an [`ITokenizer`](https://github.com/Borewit/strtok3#tokenizer) source.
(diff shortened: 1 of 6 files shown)


core.d.ts
```diff
@@ -31,19 +31,21 @@ The file type is detected by checking the [magic number](https://en.wikipedia.or
 If file access is available, it is recommended to use `.fromFile()` instead.
 
 @param buffer - An Uint8Array or ArrayBuffer representing file data. It works best if the buffer contains the entire file. It may work with a smaller portion as well.
+@param options - Options to override default behaviour.
```
sindresorhus, line 34:
```suggestion
@param options - Options to override default behavior.
```

index.test-d.ts
```diff
@@ -13,19 +16,35 @@ import {
...
+// FileTypeFromBlob
+(async () => {
+	const result = await fileTypeFromBlob(new Blob());
+	if (result !== undefined) {
+		expectType<string>(result.ext);
+		expectType<string>(result.mime);
+	}
+
+	expectType<FileTypeResult | undefined>(await fileTypeFromBlob(new Blob(), {customDetectors: []}));
+})();
```
sindresorhus, line 45: We don't need to duplicate tests for each method. Generally, type tests are only useful if they do something more than just check what we can already see from the types. Same with the rest of the changes in this file.

readme.md
```diff
@@ -130,13 +130,17 @@ Type: `Uint8Array | ArrayBuffer`
 
 A buffer representing file data. It works best if the buffer contains the entire file. It may work with a smaller portion as well.
 
-### fileTypeFromFile(filePath)
+#### options
+
+See [file type-options](#file-type-options).
```
sindresorhus, line 135: We don't need this for every method. It's implied that it's the options from the options section.

readme.md
```diff
@@ -341,6 +359,19 @@ Returns a `Set<string>` of supported file extensions.
 
 Returns a `Set<string>` of supported MIME types.
 
+### File-type options
+The following optional options are supported:
```
sindresorhus, line 363:
```suggestion
### Options

The following optional options are supported:
```

readme.md
```diff
@@ -341,6 +359,19 @@ Returns a `Set<string>` of supported file extensions.
...
+
+- `customDetectors`: Array of custom file type detectors, to run before default detectors.
```
sindresorhus, line 365:
```suggestion
#### customDetectors

Array of custom file type detectors, to run before default detectors.

```

core.d.ts
```diff
@@ -24,35 +24,38 @@ export type FileTypeResult = {
 };
 
 /**
-Detect the file type of a `Uint8Array`, or `ArrayBuffer`.
+Detect the file type of `Uint8Array`, or `ArrayBuffer`.
```
sindresorhus, line 27: Don't do unrelated changes. `a` was correct.

```suggestion
Detect the file type of a `Uint8Array` or `ArrayBuffer`.
```

readme.md
```diff
@@ -341,6 +343,23 @@ Returns a `Set<string>` of supported file extensions.
 
 Returns a `Set<string>` of supported MIME types.
 
+### File-type options
```
sindresorhus, line 346:
```suggestion
### Option
```

readme.md
```diff
@@ -341,6 +343,23 @@ Returns a `Set<string>` of supported file extensions.
...
+
+The following optional options are supported:
```
sindresorhus, line 348: Moot

readme.md
```diff
@@ -341,6 +343,23 @@ Returns a `Set<string>` of supported file extensions.
...
+
+#### customDetectors
+
+Array of custom file type detectors, to run before default detectors.
+
+For example:
```
sindresorhus, line 354:
```suggestion
Array of custom file type detectors to run before default detectors.

For example:

```

readme.md
```diff
@@ -353,8 +372,9 @@ Detectors can be added via the constructor options or by directly modifying `Fil
 
 ### Example adding a detector
 
+For example:
```
sindresorhus, line 375:
```suggestion
For example:

```

index.test-d.ts
```diff
@@ -13,39 +16,91 @@ import {
...
+	expectType<FileTypeResult | undefined>(await fileTypeParser.fromStream(new Readable()));
+
+	// From a DOM type Blob
+	expectType<FileTypeResult | undefined>(await fileTypeParser.fromBlob(new Blob()));
+
+	// From a DOM type Web byte ReadableStream
+	expectType<FileTypeResult | undefined>(await fileTypeParser.fromStream(new ReadableStream({type: 'bytes'})));
+
+	// From a Node type byte ReadableStream
+	expectType<FileTypeResult | undefined>(await fileTypeParser.fromStream(new NodeReadableStream({type: 'bytes'})));
```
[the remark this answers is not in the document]
sindresorhus: Type tests can be useful for advanced types, but just testing what it accepts is simply duplicating the type definition. index.test-d.ts was a mistake in general. Most of the time, just getting TS to validate the index.d.ts is enough.

readme.md
```diff
@@ -304,6 +304,8 @@ Type: [`stream.Readable`](https://nodejs.org/api/stream.html#stream_class_stream
 
 Type: `object`
 
+May also include [generic file-type options](#options).
```
sindresorhus, line 307:
```suggestion
Supports the following options in addition to the [general options](#options):
```

sindresorhus (changes requested):

index.test-d.ts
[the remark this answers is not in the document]
Borewit: I tried to make it more effective, and got across an issue of using mixed options (FileTypeOptions and StreamOptions), lacking right declaration. Something I could have found as well by having a critical look to the TypeScript declaration, indeed.

Corrected that, as well I (believe) improved declaration of allowing mixed 2f6aad8777ee36c8d82b927e074ab19849535d65, apparently an intersection works better. From a mathematical definition this does not make sense to me, but in the TypeScript world this seems to better reflect the type intended.

If there are things in ndex.test-d.ts, which you believe are moot, please just toss them out. For me it to arbitrary what should be in or out. I believe it is now in a more sensible and consistent state, as when found it prior to this PR.

Issue #760, opened 2025-05-22
Current tar file detection is unreliable


Borewit:
### Description

The current tar file detection does not detect every tar file archive.

### Steps to reproduce: 

Download https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.7.12.tar.gz

```bash
gunzip linux-6.7.12.tar.gz
```

```js
import { fileTypeFromFile } from 'file-type'

const result = await fileTypeFromFile("linux-6.7.12.tar");
console.log(JSON.stringify(result));
```

Outputs
```
undefined
```

This issue seems to be related to **some** [pax](https://en.wikipedia.org/wiki/Pax_(command)) tar archives, where the first, and only the (few) first 512 tar blocks(s), do not seem to have a valid checksum set. 

### Existing Issue Check

- [x] I have searched the existing issues and could not find any related to my problem.

### ESM (ECMAScript Module) Requirement Acknowledgment

- [x] My project is an [ESM project](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) and my `package.json` contains the following entry: `"type": "module"`.

### File-Type Scope Acknowledgment

- [x] I understand that file-type detects binary file types and not text or other formats.

Related:
- #76

PR #762, opened 2025-05-23
Fix detection of some PAX TAR formats

Borewit:
Resolves: #760


Borewit:
Just in case we may conclude we need this in the future, this is an updated version the TAR checksum algorithm able to detect this kind of TAR files. This needs to read multiple blocks. Problem is that I have no way to do recursive reads yet, so cannot be used in a detection easily. 

```ts
import { StringType } from 'token-types';
import { type ITokenizer } from 'strtok3';

export async function isTarHeaderChecksumMatches(tokenizer: ITokenizer): Promise<boolean> {
  const blockSize = 512;
  const typeflagOffset = 156;
  let blockIndex = 0;

  while (true) {
    const header = new Uint8Array(blockSize);
    const size = await tokenizer.peekBuffer(header, { mayBeLess: true });

    console.debug(`Reading block at index ${blockIndex}, bytes read: ${size}`);

    if (size < blockSize) {
      console.debug('Less than 512 bytes read — exiting.');
      break;
    }

    if (header.every(byte => byte === 0)) {
      console.debug('Encountered empty block — assuming end of archive.');
      break;
    }

    const typeflag = String.fromCharCode(header[typeflagOffset]);
    console.debug(`typeflag: "${typeflag}"`);

    const rawSumStr = new StringType(8, 'ascii').get(header, 148);
    const cleanedSumStr = rawSumStr.replace(/\0.*$/, '').trim();

    if (cleanedSumStr === '') {
      console.debug('Checksum field empty or spaces — skipping block.');
      await tokenizer.ignore(blockSize);
      blockIndex++;
      continue;
    }

    const readSum = Number.parseInt(cleanedSumStr, 8);

    console.debug(`Raw checksum string: "${rawSumStr}", cleaned: "${cleanedSumStr}", parsed: ${readSum}`);

    if (Number.isNaN(readSum)) {
      console.debug('Checksum is NaN — skipping block.');
      await tokenizer.ignore(blockSize);
      blockIndex++;
      continue;
    }

    let sum = 0;
    for (let i = 0; i < blockSize; i++) {
      sum += (i >= 148 && i < 156) ? 0x20 : header[i];
    }

    console.debug(`Calculated checksum: ${sum}`);

    if (readSum === sum) {
      console.debug('Checksum valid.');
      if (typeflag !== 'g' && typeflag !== 'x') {
        return true;  // Found valid regular header
      } else {
        console.debug(`Skipping extended header of type "${typeflag}".`);
      }
    } else {
      console.debug('Checksum mismatch — skipping block.');
    }

    await tokenizer.ignore(blockSize);
    blockIndex++;
  }

  return false;
}
```

PR #764, opened 2025-06-09
Add UTF-16 (BE and LE) string compare

Borewit:
Add internal support to compare UTF16-BE and UTF16-LE strings.

Reason I added this, is because I want to follow up with support Windows registry change files, which are prefixed with `Windows Registry Editor Version 5.00`,  in a UTF16-LE encoded, BOM prefixed text file.

Keeps the code more compact and readable, if we use the actual strings we are comparing.

### DIFF
--- a/test.js
+++ b/test.js
@@ -958,6 +958,10 @@ test('should detect MPEG frame which is out of sync with the mpegOffsetTolerance
 	t.deepEqual(result, {ext: 'mp3', mime: 'audio/mpeg'}, 'detect an MP3 which 1 byte out of sync');
 });
 
+function loopEncoding(t, stringValue, encoding) {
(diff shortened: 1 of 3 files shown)


test.js
```diff
@@ -10,6 +10,7 @@ import {Parser as ReadmeParser} from 'commonmark';
 import * as strtok3 from 'strtok3/core';
 import {areUint8ArraysEqual} from 'uint8array-extras';
 import {getStreamAsArrayBuffer} from 'get-stream';
+import {stringToBytes} from './util.js'; // Adjust path as needed
```
sindresorhus, line 13:
```suggestion
import {stringToBytes} from './util.js';
```

sindresorhus (changes requested):

PR #767, opened 2025-06-09
Add support for Windows registry Hives file (`.dat`)

### LINKED ISSUE #429: Support Windows Registry file
> [Windows Registry](https://en.wikipedia.org/wiki/Windows_Registry) file

Filename extension: `dat`
Media type: `zz-application/zz-winassoc-dat` (or `application/octet-stream`) [[1](https://www.whatisfileextension.com/dat/)] [[2](https://www.whatisfile.com/dat)]

Magic numbers: `72 65 67 66`

Borewit:
- Add extension: `.dat` (not very deterministic)
- Add MIME-type: `application/x-ft-windows-registry-hive` (invented by us, therefor prefixed with `x-ft-`)

Resolves #429


core.js
```diff
@@ -947,6 +947,13 @@ export class FileTypeParser {
 			};
 		}
 
+		if (this.checkString('regf')) {
+			return {
+				ext: 'dat',
+				mime: 'application/x-windows-registry-hive', // Invented by us
```
sindresorhus, line 953: I wonder if we should start prefix types when we invent them? To make it clear.

```suggestion
				mime: 'application/x-ft-windows-registry-hive', // Invented by us
```

readme.md
```diff
@@ -494,6 +494,7 @@ abortController.abort(); // Abort file-type reading from the Blob stream.
 - [`cr3`](https://fileinfo.com/extension/cr3) - Canon Raw image file (v3)
 - [`crx`](https://developer.chrome.com/extensions/crx) - Google Chrome extension
 - [`cur`](https://en.wikipedia.org/wiki/ICO_(file_format)) - Icon file
+- [`dat`](https://en.wikipedia.org/wiki/Windows_Registry) - Windows registry hives file
```
sindresorhus, line 497:
```suggestion
- [`dat`](https://en.wikipedia.org/wiki/Windows_Registry) - Windows registry hive file
```

since a file only has a single hive

core.js
Borewit: Although `application/x-lz4` is marked is _invented by us_, I propose to keep that one as is, as that one if actually widely used. So not totally invented by us in reality. Probably best to just remove the _invented by us_ comment.

sindresorhus: > Probably best to just remove the invented by us comment.

👍

PR #773, opened 2025-08-06
Handle partial unzip

avoylenko:
Do not throw the `EndOfStreamError()` exception in case you pass only partial zip data.

Code to reproduce the error:
```
file.stream.once('data', async (firstChunk) => {
    const type = await fileTypeFromBuffer(firstChunk)
})
```

Stack trace:
```
EndOfStreamError: End-Of-Stream
at BufferTokenizer.peekBuffer (file:///opt/novatalks.engine/node_modules/strtok3/lib/BufferTokenizer.js:38:19)
at BufferTokenizer.readBuffer (file:///opt/novatalks.engine/node_modules/strtok3/lib/BufferTokenizer.js:24:38)
at BufferTokenizer.readToken (file:///opt/novatalks.engine/node_modules/strtok3/lib/AbstractTokenizer.js:32:32)
at ZipHandler.unzip (file:///opt/novatalks.engine/node_modules/@tokenizer/inflate/lib/index.js:127:61)
at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
at async Object.detectConfident [as detect] (file:///opt/novatalks.engine/node_modules/file-type/core.js:521:4)
at async FileTypeParser.fromTokenizer (file:///opt/novatalks.engine/node_modules/file-type/core.js:211:21)
```

PR #774, opened 2025-08-11
Handle zip errors so we can still tell the file is a zip.

MustafaEminn:
In the web environment, fileTypeFromStream often gives End-Of-Stream errors with ZIP files. Even if the ZIP parser fails, we should still return that the file is a ZIP.

### DIFF
--- a/core.js
+++ b/core.js
@@ -601,7 +601,6 @@ export class FileTypeParser {
 					}
 				});
 			} catch {}
-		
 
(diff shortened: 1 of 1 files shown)


avoylenko:
Seems like https://github.com/sindresorhus/file-type/pull/773

MustafaEminn:
> Seems like #773

Ah you are right. I will follow #773

PR #779, opened 2025-10-20
Add support for Mach-O Universal (aka "Fat") binaries and additional architectures

samumbach:
## Mach-O single architecture (aka "thin") binaries

Mach-O "thin" binaries have different magic numbers depending on the architecture "bitness" and endianness (source: https://en.wikipedia.org/wiki/Mach-O#Mach-O_header):

- `0xFE 0xED 0xFA 0xCE` 32-bit, big-endian (e.g., ppc)
- `0xFE 0xED 0xFA 0xCF` 64-bit, big-endian (e.g., ppc64)
- `0xCE 0xFA 0xED 0xFE` 32-bit, little-endian (e.g., i386)
- `0xCF 0xFA 0xED 0xFE` 64-bit, little-endian (e.g., x86_64 and arm64)

## Mach-O multi-architecture (aka Universal or "fat") binaries

Mach-O multi-architecture binaries, unfortunately, have the same `0xca 0xfe 0xba 0xbe` magic number as Java `.class` files:

- https://en.wikipedia.org/wiki/Mach-O#Multi-architecture_binaries
- https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html

These two formats can be disambiguated by the next 4 bytes, which specify the number of binaries in a Mach-O multi-architecture binary (should be a small-ish number) or the Java Class format minor and major version (first supported major version was 45). We can adopt a similar approach to that used by `file`: https://github.com/file/file/blob/master/magic/Magdir/cafebabe

## See also:

- #615 
- #591


samumbach:
NOTE: I was unable to build or find a reasonably sized and redistributable ppc64 binary. The only one I could find is `Chess.app` from Mac OS X 10.5 Leopard. The ppc64 binary is 245KB and I'm unsure if the licensing allows it to be added to this git repo.

Borewit (approved):
LGTM

PR #787, opened 2025-12-21
Add support for SPSS data files

madflow:
* Adds support for the SPSS `.sav` data/system file format.
* Ported from libmagic https://github.com/file/file/blob/9ed4d8c65854d9d28519291f21dd92c44c4abc18/magic/Magdir/macintosh#L394
* https://www.loc.gov/preservation/digital/formats/fdd/fdd000469.shtml
* https://www.gnu.org/software/pspp/pspp-dev/html_node/System-File-Format.html
* The `application/x-spss-sav` mimetype is not "standardized" - but used in software like [Wolfram](https://reference.wolfram.com/language/ref/format/SAV.html) or [Apache Tika](https://ciam-tika.straumann.com/mime-types/application/x-spss-sav) It needs to be so specific since this library relies on a specific mime-type.

### DIFF
--- a/core.js
+++ b/core.js
@@ -543,14 +543,6 @@ export class FileTypeParser {
 			};
 		}
 
-		// SPSS Statistical Data File
-		if (this.checkString('$FL2') || this.checkString('$FL3')) {
-			return {
-				ext: 'sav',
-				mime: 'application/x-spss-sav',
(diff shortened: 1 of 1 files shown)


Borewit (approved):
LGTM

PR #796, opened 2026-09-10
Add support for ISO 9660 (.iso)

abhiyadav79:
Adds detection for ISO 9660 optical disc images (`.iso`).

## Format reference
- Spec / overview: https://en.wikipedia.org/wiki/ISO_9660

## Signature
The Primary Volume Descriptor lives in sector 16 (2048-byte sectors), and its first field after the type code is the 5-byte standard identifier `CD001` — at absolute byte offset 32769. This is the same signature used by `file(1)`/libmagic to identify ISO 9660 images.

## MIME type
`application/x-iso9660-image` — the de facto MIME type returned by the `file` command and used by shared-mime-info, since ISO 9660 has no IANA-registered media type.

## Notes
Since the signature sits well past the default detection sample size, this check peeks at an absolute stream position (via `mayBeLess`) rather than the shared buffer used by other checks, and is gated on `tokenizer.position` not having already advanced past that offset (some earlier checks, e.g. ID3v2 tag skipping, can advance the position on non-seekable streams, where peeking backward is not possible).

`npm test` passes (1567/1567).


sindresorhus:
Could we limit this to random-access tokenizers?

Because there is no prefix to gate the ISO check, every otherwise-unrecognized stream now has to produce 32,774 bytes or reach EOF. I manually verified this with a stream that emits 1,000 bytes and stays open: main returns undefined immediately, while this branch remains pending until it is aborted. That seems like too broad a behavior change for adding one format.

I think ISO support itself makes sense, but the simpler tradeoff is to support it for random-access inputs only. That still covers files, buffers, blobs, and fileTypeStream() when the caller uses sampleSize: 32_774, without delaying every unknown stream. Direct unknown-size stream detection can be documented as unsupported here. Please add a test for the explicit sampleSize; the generated stream test currently only checks that the bytes are preserved and doesnt check the detected type.

Also, the comment should say the Volume Descriptor Set starts at sector 16. The Primary Volume Descriptor is not necessarily the first descriptor, although the CD001 check itself is correct.

abhiyadav79:
Good catch, thanks. Pushed a fix:

- Gated the check on `tokenizer.supportsRandomAccess()` instead of size/position bookkeeping, so it only runs for buffers/files/blobs (and `fileTypeStream()` when the caller passes a large enough `sampleSize`). Unrecognized unknown-size streams now resolve immediately again — verified manually with a stream that emits 1000 bytes and stays open, it now returns `undefined` right away instead of hanging.
- Added a `fileTypeStream()` test that asserts the detected type with `sampleSize: fixture.length`, not just byte preservation.
- Reworded the comment: the Volume Descriptor *Set* starts at sector 16, and the Primary Volume Descriptor isn't necessarily first (every descriptor carries `CD001` at the same offset regardless of type, so checking the first one is still sufficient).

PR #800, opened 2026-09-15
Read the ZIP central directory for random-access input

### LINKED ISSUE #799: Streamed (data-descriptor) `.docx` is detected as `application/zip`
### Description

> [!NOTE]
> This issue was researched and written by [Claude Code](https://claude.com/claude-code).
> The reproduction below was run and verified against `file-type@22.1.0` before posting.

A `.docx` written to a non-seekable stream — every Google Docs export, among others — is
detected as `application/zip` rather than `docx`, as of `file-type@22.1.0`.

**Why it happens**

A zip written to a stream can't seek back to patch each local file header, so it sets bit 3
of the general-purpose flags and writes zeroes for the CRC and both sizes, putting the real
values in a data descriptor *after* the entry's data. `file-type` therefore has to scan
forward for the descriptor signature to learn where each entry ends, and bounds that scan
with `maximumUntrustedSkipSizeInBytes`.

OOXML producers are free to order the archive as they like, and Google Docs puts
`[Content_Types].xml` **last**. In a real export the embedded images sit in front of it, so
the skip budget is exhausted several MB before the parser reaches the one entry that
identifies the format.

The directory-name fallback doesn't rescue it either, because
`getOpenXmlFileTypeFromZipEntries` is gated on `hasContentTypesEntry`:

```js
if (
    !openXmlState.hasContentTypesEntry
    || openXmlState.hasUnparseableContentTypes
    || openXmlState.isParsingContentTypes
    || openXmlState.hasParsedContentTypesEntry
) {
    return;
}

return getOpenXmlFileTypeFromDirectoryNames(openXmlState);
```

When the scan stops early, `[Content_Types].xml` is never *seen*, so `hasContentTypesEntry`
stays `false` and the `word/` heuristic is skipped — even though `hasWordDirectory` is
already `true`. The result is a bare `zip`.

This isn't specific to Word: `xlsx` and `pptx` fail the same way, and I think #688 is the
same underlying case, reported as slowness rather than misdetection.

**Reproduction**

Self-contained, no fixtures or dependencies beyond `file-type@22.1.0`. It writes the same
four parts twice, once as an ordinary zip and once as a streamed one. Also attached as a
runnable file: https://gist.github.com/alecgibson/edfd089a9ebdffd711b2f46027666d90

```console
npm install file-type@22.1.0    # in a package.json with "type": "module"
node file-type-streamed-docx-repro.js
```

```js
import {createHash} from 'node:crypto';
import {crc32, deflateRawSync} from 'node:zlib';
import {writeFileSync, rmSync} from 'node:fs';
import {fileTypeFromFile} from 'file-type';

function buildZip(parts, {streamed}) {
	const chunks = [];
	const central = [];
	let offset = 0;

	const push = (buffer) => {
		chunks.push(buffer);
		offset += buffer.length;
	};

	for (const {name, contents} of parts) {
		const filename = Buffer.from(name);
		const compressed = deflateRawSync(contents);
		const checksum = crc32(contents);
		const localOffset = offset;

		const local = Buffer.alloc(30);
		local.writeUInt32LE(0x04_03_4b_50, 0);
		local.writeUInt16LE(20, 4);
		local.writeUInt16LE(streamed ? 0x08 : 0, 6);
		local.writeUInt16LE(8, 8);
		if (!streamed) {
			local.writeUInt32LE(checksum, 14);
			local.writeUInt32LE(compressed.length, 18);
			local.writeUInt32LE(contents.length, 22);
		}

		local.writeUInt16LE(filename.length, 26);
		push(local);
		push(filename);
		push(compressed);

		if (streamed) {
			const descriptor = Buffer.alloc(16);
			descriptor.writeUInt32LE(0x08_07_4b_50, 0);
			descriptor.writeUInt32LE(checksum, 4);
			descriptor.writeUInt32LE(compressed.length, 8);
			descriptor.writeUInt32LE(contents.length, 12);
			push(descriptor);
		}

		const header = Buffer.alloc(46);
		header.writeUInt32LE(0x02_01_4b_50, 0);
		header.writeUInt16LE(20, 4);
		header.writeUInt16LE(20, 6);
		header.writeUInt16LE(streamed ? 0x08 : 0, 8);
		header.writeUInt16LE(8, 10);
		header.writeUInt32LE(checksum, 16);
		header.writeUInt32LE(compressed.length, 20);
		header.writeUInt32LE(contents.length, 24);
		header.writeUInt16LE(filename.length, 28);
		header.writeUInt32LE(localOffset, 42);
		central.push(Buffer.concat([header, filename]));
	}

	const centralOffset = offset;
	for (const entry of central) push(entry);

	const end = Buffer.alloc(22);
	end.writeUInt32LE(0x06_05_4b_50, 0);
	end.writeUInt16LE(central.length, 8);
	end.writeUInt16LE(central.length, 10);
	end.writeUInt32LE(offset - centralOffset, 12);
	end.writeUInt32LE(centralOffset, 16);
	push(end);

	return Buffer.concat(chunks);
}

// Incompressible, so it counts in full against the scan budget — real exports reach this
// with embedded images.
function filler(bytes) {
	const chunks = [];
	let block = createHash('sha256').update('seed').digest();
	for (let written = 0; written < bytes; written += block.length) {
		chunks.push(block);
		block = createHash('sha256').update(block).digest();
	}

	return Buffer.concat(chunks, bytes);
}

// `[Content_Types].xml` goes last, as it does in Google Docs' exports.
const parts = [
	{name: 'word/document.xml', contents: Buffer.from('<w:document/>')},
	{name: '_rels/.rels', contents: Buffer.from('<Relationships/>')},
	{name: 'word/media/image1.bin', contents: filler(3 * 1024 * 1024)},
	{
		name: '[Content_Types].xml',
		contents: Buffer.from(
			'<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
			+ '<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">'
			+ '<Override PartName="/word/document.xml" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>'
			+ '</Types>',
		),
	},
];

for (const streamed of [false, true]) {
	const path = `./fixture-${streamed ? 'streamed' : 'regular'}.docx`;
	writeFileSync(path, buildZip(parts, {streamed}));
	console.log(`${streamed ? 'streamed' : 'regular '} zip ->`, await fileTypeFromFile(path));
	rmSync(path);
}
```

Actual, on `file-type@22.1.0` / Node 24.14.1:

```
regular  zip -> { ext: 'docx', mime: 'application/vnd.openxmlformats-officedocument...' }
streamed zip -> { ext: 'zip', mime: 'application/zip' }
```

Expected: both report `docx`. The two files contain identical entries, in identical order,
and differ only in how the sizes are recorded.

**Possible fix**

When an entry's local header declares unknown sizes, the archive's central directory — at
the end of the file — already holds the real sizes and the complete entry list, so no
forward scanning is needed. Reading it would give an exact answer for the whole OOXML
family (`docx`/`docm`/`dotx`/`dotm` and the Excel and PowerPoint equivalents) instead of a
directory-name guess, and would fix `xlsx`/`pptx` in the same change.

It does need seekable input, so it would apply to `fileTypeFromFile` and `fromBuffer` but
not to a pure stream — which seems like the right trade, since the current behaviour for
these files is simply wrong rather than merely imprecise.

### Existing Issue Check

- [x] I have searched the existing issues and could not find any related to my problem.

- [x] My project is an ESM project and my `package.json` contains the following entry: `"type": "module"`.

- [x] I understand that file-type detects binary file types and not text or other formats.

alecgibson:
Fixes #799

A `.docx` written to a non-seekable stream — every Google Docs export — is detected as `zip`. Such an archive leaves the sizes out of its local file headers and defers them to a data descriptor after each entry, so reaching `[Content_Types].xml`, which Google Docs writes last, means scanning forward past the embedded media, and that scan is bounded to 1 MiB per entry.

Where the input can seek, the central directory at the end of the archive already records every entry's name, size and offset, so we read that instead and fall back to the scan whenever it is absent, ZIP64, inconsistent or over the existing limits. This routes all random-access ZIP detection through the directory rather than only streamed archives, and once the directory is trusted there is no retry — so an archive whose directory misplaces an entry now reports `zip` where the scan would have detected it.

The central-directory reader is reimplemented rather than reused from `@tokenizer/inflate` because upstream offers no injection point for the bounds this needs and doesn't export its end-of-central-directory token.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>


alecgibson:
Sorry for the noise 🙈 this was a way bigger change than I anticipated. Closing and looking at it again locally hopefully with a smaller diff

